PR #11, opened 2013-04-06
TeamCity improvements

alastairs:
Display the currently-executing psake task in TeamCity.


alastairs:
:warning: **DO NOT MERGE THIS** :warning: 

It fires up a PowerShell file in notepad, and then causes subsequent builds to fail because notepad retains a lock on the build directory. No idea WTF is going on here.

alastairs:
I'm a bit confused why TeamCity hasn't updated this pull request, but it's now green and ok to be merged.

Issue #1, opened 2013-04-04
Ordering of the login methods in inconsistent


adrianbanks:
In the menu, the order goes GitHub, Twitter, Google. In the submit session page, it goes Twitter, Google, GitHub. Probably should be the same.

alastairs:
I was going to try and fix this now if it was trivial, but I've looked in about four different files and haven't found the appropriate bit for the form displayed on the session submission page yet.

...I am using vim, on a Mac :smile:

FWIW, I deliberately re-ordered the login providers in the menu, in the hope that it might encourage people to use GitHub over Twitter or Google (and then Twitter over Google).  The order they were in before was the order we added support for them, which was as good as random order.

adrianbanks:
These are added in `AuthConfig.cs`. I think we can just reorder them there to match the order in the menu (although we are then reliant on the ordering being kept in the auto provider code).

alastairs:
Hmmm, seems to me like the order in which they are rendered should be separate from the order in which they are configured... I'll take a look over the weekend, and at #2 and #3 too.  There's a whole bunch of styling that needs doing on the authorised pages right now.

adrianbanks:
Yes, I agree but for a quick fix we could just reorder them.

The page you get when you click submit a session is particularly horrible. I did some playing with it last night to make it nicer so already have some ideas.

adrianbanks:
https://github.com/dddeastanglia/DDDEastAnglia/pull/13

PR #13, opened 2013-04-06
Issue #1: Ordering of the login methods is inconsistent

adrianbanks:
I've explicitly output the buttons, as there is no way to order them using the model approach as they come out of the OAuth provider and I cannot change that.

I've also added the little icons to match the menu.

### DIFF
--- a/DDDEastAnglia/Views/Account/Login.cshtml
+++ b/DDDEastAnglia/Views/Account/Login.cshtml
@@ -32,7 +32,7 @@
                 @Html.LabelFor(m => m.RememberMe, new { @class = "checkbox" })
             </li>
         </ol>
-        <input type="submit" value="Login" />
+        <input type="submit" value="Log in" />
     </fieldset>
     <p>
         @Html.ActionLink("Register", "Register") if you don't have an account.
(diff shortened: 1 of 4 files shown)


DDDEastAnglia/Views/Account/Login.cshtml
```diff
@@ -32,7 +32,7 @@
                 @Html.LabelFor(m => m.RememberMe, new { @class = "checkbox" })
             </li>
         </ol>
-        <input type="submit" value="Log in" />
+        <input type="submit" value="Login" />
```
alastairs: This was correct before. "Login" == noun, "Log in" == verb.

adrianbanks: I did think that, but you said that you had explicitly done the menu the way it was, so I made it match that. I can change them both to "log in".

alastairs: Yes please: my commit 766960ff set the menu button to read "Log In", and it looks to be that way on the live site, but perhaps it's got overwritten somewhere since.

(No need to capitalise the i of "in" on this button.)

alastairs:
Fixes #1.

Issue #16, opened 2013-04-07
The tweet link logs you out of Twitter


adrianbanks:
When you click the tweet link below a session, you are asked to log into twitter, even if you were already logged in.

alastairs:
:scream:

adrianbanks:
https://github.com/dddeastanglia/DDDEastAnglia/pull/17

alastairs:
I wonder if this is why no-one has used this feature yet...

adrianbanks:
Possibly. I intend on going through the whole sign-up & submit process with my ux hat on tomorrow. It will hopefully give it a thorough testing and flush out some bugs, but should also pick up on anything like this - it is possible, but hard or not worth the hassle.

alastairs:
Fixed by #17.

PR #17, opened 2013-04-07
Issue #16: The tweet link logs you out of Twitter

adrianbanks:
Change the twitter links so that it doesn't log the user out of twitter.

### DIFF
--- a/DDDEastAnglia/Views/Session/Details.cshtml
+++ b/DDDEastAnglia/Views/Session/Details.cshtml
@@ -15,11 +15,11 @@
 <div class="tweet-session">
 @if (User.Identity.Name == Model.SpeakerUserName)
 {
-    <a href="https://twitter.com/share" class="twitter-share-button" data-url="@Request.Url" data-text="Check out my session for #dddea @Model.SessionTitle @Request.Url" data-count="none">Tweet</a>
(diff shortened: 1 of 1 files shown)


DDDEastAnglia/Views/Session/Details.cshtml
```diff
@@ -15,11 +15,11 @@
 <div class="tweet-session">
 @if (User.Identity.Name == Model.SpeakerUserName)
 {
-    <a href="https://twitter.com/share" class="twitter-share-button" data-url="@Request.Url" data-text="Check out my session for #dddea @Model.SessionTitle @Request.Url" data-count="none">Tweet</a>
+    <a href="https://twitter.com/intent/tweet?text=@Html.Encode("Check out my session for #dddea -") @Model.SessionTitle @Request.Url" class="twitter-share-button" data-url="@Request.Url" data-count="none">Tweet</a>
```
alastairs, line 18: I suspect that with the change to the href, none of the data-\* attributes, nor the JavaScript snippet at the bottom, are required any longer. We could also style the button using Bootstrap's btn class (and, perhaps, btn-primary, although that's looking pretty ugly at the moment with DDDEA green), and use the nice Font Awesome Twitter icon.

I haven't seen any of this running live, though, I'm just eyeballing HTML, so take it all with a pinch of salt :grin:

adrianbanks, line 18: There is some twitter js that styles the button to look like the standard twitter button that you see everywhere. The data-\* attribute are used by this to make the style.

alastairs, line 18: Hmmm, ok. :+1:

PR #18, opened 2013-04-07
Add an SSDT database project

alastairs:
:alarm_clock: :warning: **PLEASE READ** :warning: :alarm_clock: 

After this pull request is merged, you will need SQL Server Data Tools (SSDT) installed to open the database project.  SSDT is included by default in Visual Studio 2012, and is supported by Visual Studio 2010 _with SP1_; the latest version (December 2012 update) is available at http://msdn.microsoft.com/en-us/data/hh297027.

The SSDT project automatically deploys to a LocalDB database, so you will probably need SQL Server 2012 with LocalDB installed too. If you're using this project, you no longer need the .mdf and .ldf files that we've been sharing on Dropbox.

This is stage one of getting database deployments going.


alastairs:
This is good to merge once someone's taken a look at it.

alastairs:
Rebased onto latest `production` branch.

alastairs:
Rebased again.

alastairs:
I've not heard any objections, so I'm going to merge this now. I hope everyone has the necessary bits installed.

adrianbanks:
This doesn't make a `_MigrationHistory` table. Is that going to be a problem?

alastairs:
No, Entity Framework creates that itself when it does the first migration on the database.

adrianbanks:
Are you sure? Didn't Ian (or possibly Phil) have an issue where something broke because they table wasn't present?

alastairs:
Not that I remember (although that's not to say it's not the case!). It seemed to work ok for me when I tested it out.

alastairs:
I have pulled this out of the staging branch because there's a problem with the SSDT binaries being left locked by MSBuild at the end of a build, which then causes all future builds to fail on that agent.

PR #20, opened 2013-04-07
Fix #19: visual separation of speakers.

alastairs:
There are a few changes in here, I got a bit carried away. Basically,
the speakers page looks a bit nicer :smile:  An overview of the changes:
- Remove the inline headings in favour of Font Awesome icons where
  appropriate (i.e., a speaker's links).
- Delineate the speaker's bio by separating it visually, slightly: no
  need for the "Bio:" heading.
- Make the submitted sessions an h4, and list out the sessions in a ul.
- Add 30px of space between speakers.

I'll add before and after screenshots to #19 so the two can be compared.

PR #32, opened 2013-04-10
Changes following the UX review

adrianbanks:
This is not yet complete, but I'd like to get some feedback. Things still left to do:
- [x] Require the user to enter a name and email address (or capture them) when logging in with OAuth as a primary login.

I think this can be classed as done. These others can be done separately: 
- Create a page explaining the DDD process (submission, voting, etc...) and link to it from a "New to DDD?" link on the home page
- Prompt for confirmation (jquery dialog) when deleting a session.


alastairs:
I'm confused: the last one of your "to-do" items is the very thing I fixed in #20 (see [the staging site](http://dddeastaging.azurewebsites.net/Speaker)). 

Here's some things I've noticed:
- The "Log in" link doesn't fit with the rest of the drop-down menu. It needs an icon (perhaps `icon-user`?), and the text perhaps needs clarifying a little: "Log in with local account"/"Log in with DDD EA account"?
- In the drop-down menu, the Twitter icon should be black like the others.
- Logging in always redirects you to the homepage. This is annoying when you've clicked a "submit a session" link whilst unauthenticated, as you expect to be taken to the session submissions page after authenticating.
- The footer would be better laid out in a line (e.g., Copyright -- Email -- Twitter), rather than in three rows.
- The banner that you've added is, in its current form, annoying. It's too big, loud and shouty to be permanently displayed, so it should be possible to dismiss it. It should not form part of the body text the way it currently does; it might be better if it hung off the navigation bar instead.
- The colour of "DDD East Anglia" in the banner is a shade of green that doesn't appear anywhere else on the site. It would probably be better to leave this as the default body colour.  
- The Sessions' page should:
  - include the speaker's Gravatar, not the `icon-user` icon.
  - the Tweet button should be better integrated into the flow, ideally after the abstract of the session. The currently placing of it de-emphasises it as an action, which is the exact opposite of what we want. It could perhaps be improved to say "Tweet this session" as well.
  - not include the "Please provide the details of your session below" text; this is redundant, particularly at the size which it is currently displayed.
- All the borders on the Sessions' page add too much visual clutter. The page in general should be more consistent with the Speakers' page.
- The Edit Profile page still needs a fair bit of work:
  - Required fields (name, email address) should be identified as such
  - "Please provide as much information as possible" is perhaps a nice line to include, but not as the form title.
  - The new HTML5 input types would be useful to use on the Email, Mobile Phone and Website fields.
- The markdown preview areas could perhaps do with a simple "Preview" title.
- Field labels should be bold, and, perhaps in-line in the field itself. The `placeholder` attribute will allow you to accomplish in-line labels on HTML5-compatible browsers; you can use a Modernizr polyfill with a jQuery plugin to achieve this in incompatible browsers (but, given our audience, I think we can safely target HTML5-capable browsers only). 
- Clicking the "Abstract" label does not focus the Abstract text area; ditto the "Bio" link on the Edit Profile page.
- Buttons should be styled with the `btn` class. The primary action button should be additionally styled with the `btn-primary` class. 
- "Cancel" links should be buttons, not links.
- Contacts page still needs some work.
- On the Manage Logins page, the Twitter icon should be black like the other buttons. This page appears to be unfinished: it just says "Twitter" for me (it's not even a link!) and then offers the buttons. 
- The Delete session page is _weird_ (although I see this is something sort of on your to-do list). 
  - I click "Delete" on my session, then am taken to a page that displays the session completely identically, including "Edit" and "Delete" links. If I click "Delete" here, am I going to be taken back to this page, or is the session actually going to be deleted? I want a big clear button (i.e., `btn btn-large btn-danger`) to do the delete, not an ambiguous hyperlink.
  - The "Please confirm" text is styled like a heading but it's not worded as one.
  - It should state how destructive an action this is - can a deleted session be recovered?

DDDEastAnglia/Content/Site.css
```diff
@@ -128,3 +128,133 @@ a.btn, .dropdown-menu a { text-decoration: none; }
     width: 20px;
     text-align: center;
 }
+
```
alastairs, line 131: What's with all the empty lines?

DDDEastAnglia/Content/Site.css
```diff
@@ -128,3 +128,133 @@ a.btn, .dropdown-menu a { text-decoration: none; }
...
+.banner a:hover {
+    color: greenyellow;
+}
+
+a i {
+    text-decoration: none;
+}
+
+.icon-twitter {
```
alastairs: I deliberately did not do this, because the only place the Twitter icon should be styled in this colour is on the Speakers' page. This has added the colour to the Twitter icon all over the site, which is explicitly not what I intended.

DDDEastAnglia/Content/Site.css
```diff
@@ -128,3 +128,133 @@ a.btn, .dropdown-menu a { text-decoration: none; }
...
+    color: cornflowerblue;
+}
+
+.icon-globe {
```
alastairs: Again, this style was specifically for that one element, which is why I put the styling on the element and not in the stylesheet.

DDDEastAnglia/Content/Site.css
```diff
@@ -128,3 +128,133 @@ a.btn, .dropdown-menu a { text-decoration: none; }
...
+    color: green;
+}
+
+.speaker {
```
alastairs, line 242: This style should be merged with the existing `.speaker` style when it is merged into staging.

DDDEastAnglia/AccountController.generated.cs
```diff
@@ -307,34 +307,34 @@ public override System.Web.Mvc.ActionResult Register(DDDEastAnglia.Models.Regist
...
-        partial void ManageOverride(T4MVC_System_Web_Mvc_ActionResult callInfo, DDDEastAnglia.Controllers.AccountController.ManageMessageId? message);
-
-        public override System.Web.Mvc.ActionResult Manage(DDDEastAnglia.Controllers.AccountController.ManageMessageId? message)
-        {
-            var callInfo = new T4MVC_System_Web_Mvc_ActionResult(Area, Name, ActionNames.Manage);
-            ModelUnbinderHelpers.AddRouteValues(callInfo.RouteValueDictionary, "message", message);
-            ManageOverride(callInfo, message);
-            return callInfo;
-        }
+//        public override System.Web.Mvc.ActionResult Disassociate(string provider, string providerUserId)
```
alastairs: Why is all this code commented out and not deleted? Also, it's generated code, so doesn't it get overwritten when you do a build?

DDDEastAnglia/Content/Site.css
```diff
@@ -128,3 +128,133 @@ a.btn, .dropdown-menu a { text-decoration: none; }
...
+
+#wmd-preview {
+    width: 80%;
+    border: 2px dotted #ccc;
+    padding: 5px;
+    margin-bottom: 15px;
+}
+
+.session-details-own {
+    background-color: #f9f9f9;
```
alastairs: This colour is too subtle, I hadn't even noticed it when reviewing it visually.

adrianbanks:
I've addressed the items that are stuckthrough. The other have comments inline
- ~~The "Log in" link doesn't fit with the rest of the drop-down menu. It needs an icon (perhaps `icon-user`?), and the text perhaps needs clarifying a little: "Log in with local account"/"Log in with DDD EA account"?~~
- ~~In the drop-down menu, the Twitter icon should be black like the others.~~
- Logging in always redirects you to the homepage. This is annoying when you've clicked a "submit a session" link whilst unauthenticated, as you expect to be taken to the session submissions page after authenticating.

**I cannot reproduce this - after logging in you get returned to the page that you requested.**
- ~~The footer would be better laid out in a line (e.g., Copyright -- Email -- Twitter), rather than in three rows.~~
- The banner that you've added is, in its current form, annoying. It's too big, loud and shouty to be permanently displayed, so it should be possible to dismiss it. It should not form part of the body text the way it currently does; it might be better if it hung off the navigation bar instead.

**The fact that it is shouty is intentional. It is designed to draw a visitor's attention. It is only displayed on a few select pages - home, sessions and speakers - the pages that relate directly to session submission. Hanging it off the navigation bar doesn't give the same contrast, and may make it less in-your-face.**
- ~~The colour of "DDD East Anglia" in the banner is a shade of green that doesn't appear anywhere else on the site. It would probably be better to leave this as the default body colour.~~  
- The Sessions' page should:
  - ~~include the speaker's Gravatar, not the `icon-user` icon.~~
  - ~~the Tweet button should be better integrated into the flow, ideally after the abstract of the session. The currently placing of it de-emphasises it as an action, which is the exact opposite of what we want. It could perhaps be improved to say "Tweet this session" as well.~~
  - not include the "Please provide the details of your session below" text; this is redundant, particularly at the size which it is currently displayed.

**The form requires a legend (unless we ditch the fieldset tag). There's not a lot else to say in it.**
- ~~All the borders on the Sessions' page add too much visual clutter. The page in general should be more consistent with the Speakers' page.~~
- The Edit Profile page still needs a fair bit of work:
  - Required fields (name, email address) should be identified as such

**Didn't do that before, but yes good suggestion. This needs to happen all over the site. Doesn't seem trivial to implement though.**
- "Please provide as much information as possible" is perhaps a nice line to include, but not as the form title.

**Another issue with the form requiring a legend.**
- The new HTML5 input types would be useful to use on the Email, Mobile Phone and Website fields.

**I'm just using the output from the MVC helpers, same as was already there.**
- The markdown preview areas could perhaps do with a simple "Preview" title.

**I have no control over how the preview area is output, as the markdown plugin renders it. There doesn't appear to be a  style that can be applied to add it using CSS.**
- Field labels should be bold, and, perhaps in-line in the field itself. The `placeholder` attribute will allow you to accomplish in-line labels on HTML5-compatible browsers; you can use a Modernizr polyfill with a jQuery plugin to achieve this in incompatible browsers (but, given our audience, I think we can safely target HTML5-capable browsers only). 

**Seems a lot of hassle to bold a field. They also weren't bold before my change.**
- Clicking the "Abstract" label does not focus the Abstract text area; ditto the "Bio" link on the Edit Profile page.

**Again, the markdown control interferes with this.**
- ~~Buttons should be styled with the `btn` class.~~ The primary action button should be additionally styled with the `btn-primary` class. 

**The `btn-primary` class turns all the buttons dark-green, and takes the roundedness off them. Is that correct, as it doesn't look nice?**
- "Cancel" links should be buttons, not links.

**No. There is a fair amount of usability research that says that the "default" choice should be the obvious one, so making the other choice a link immediately relegates it.**
- Contacts page still needs some work.

**Didn't change this page.**
- ~~On the Manage Logins page, the Twitter icon should be black like the other buttons.~~ This page appears to be unfinished: it just says "Twitter" for me (it's not even a link!) and then offers the buttons. 

**Those come from the database, so are hard to style with the correct icon for the provider, or hyperlink (where would they hyperlink to?). Equally the buttons are submit buttons for a form. We could possibly replace it with an image for "remove".**
- The Delete session page is _weird_ (although I see this is something sort of on your to-do list). 
  - ~~I click "Delete" on my session, then am taken to a page that displays the session completely identically, including "Edit" and "Delete" links. If I click "Delete" here, am I going to be taken back to this page, or is the session actually going to be deleted? I want a big clear button (i.e., `btn btn-large btn-danger`) to do the delete, not an ambiguous hyperlink.~~
  - The "Please confirm" text is styled like a heading but it's not worded as one.

**Another issue with the form requiring a legend.**
- ~~It should state how destructive an action this is - can a deleted session be recovered?~~

> I'm confused: the last one of your "to-do" items is the very thing I fixed in #20 (see the staging site).

There was a mention in the call yesterday that we should add something to separate the items better.

alastairs:
A further suggestion: this is kind of a large pull request. Could it be split up into separate pull requests by page?

Pull request all the things!

adrianbanks:
Not sure how easy it will be to separate, especially as a lot of the changes have crossed over in the code behind and models. I'm keen to get this in sooner, as the longer it takes, the more people are exposed to oddities and problems of the current site. The more barriers we can remove for people submitting sessions the better.

DDDEastAnglia/Content/Site.css
adrianbanks, line 131: Removed them.

DDDEastAnglia/AccountController.generated.cs
adrianbanks: Still working on this. I don't have T4MVC installed, so it didn't regenerate the controllers. To get it to compile, I just commented them out for the time being.

alastairs:
I was thinking that it would be easier to get it in if it were in smaller pull requests, as they would be easier to review, etc. The review I did earlier did most of the site, because I couldn't tell exactly what was covered in this pull request and what wasn't. It's always easier to assess the impact of a smaller change than a larger one.

adrianbanks:
Yes, I agree it would be easier, but that it not how it has panned out in this case. The problem is that each change relies on parts of the other changes, and the code behind also crosses more than one change.

alastairs:
> The fact that it is shouty is intentional. It is designed to draw a visitor's attention. It is only displayed on a few select pages - home, sessions and speakers - the pages that relate directly to session submission. Hanging it off the navigation bar doesn't give the same contrast, and may make it less in-your-face.

I get that it's intentional that it be shouty, but I think you've gone a bit overboard with the shoutiness: you might as well make it blink while you're at it :smile: You also missed the main point that I was trying to make: **it should be possible to dismiss it.** It's epically annoying having it all up in my grill when I've decided I'm not interested in it any more.

> [Forms] requires a legend (unless we ditch the fieldset tag). There's not a lot else to say in it.

The [W3 HTML 5 documentation](http://www.w3.org/html/wg/drafts/html/master/forms.html#the-legend-element) says this about the `legend` element:

> The `legend` element represents a caption for the rest of the contents of the `legend` element's parent `fieldset` element, if any.

So, like a caption to a picture or a table, it should describe the set of fields contained within the fieldset. A simpler way of thinking of it would be as a title for the form, so probably it would be best to place the current page heading text into the `legend`.

> - The new HTML5 input types would be useful to use on the Email, Mobile Phone and Website fields.
> 
> I'm just using the output from the MVC helpers, same as was already there.

Ok, but this would be a good UX improvement to make: these new input types change the keyboard on mobile devices, and give the browser more knowledge about what the field is (=> better validation of the input). You can hard-code the HTML for now if that's easier than writing your own helper.

> [Placeholder and Modernizr polyfills] seems a lot of hassle to bold a field. They also weren't bold before my change.

It's a suggested improvement over what we have already, seeing as you're improving what we have already. The placeholder attribute doesn't bold a field, it adds some text to the field that disappears when the user enters their own value. It would be particularly useful for the name and email fields on the profile page, but it can be used more generally for form labels.

I also said that actually we probably don't need the Modernizr polyfill stuff, because developers usually run the latest version of any given browser, and so they will natively support the placeholder attribute. Therefore, the hassle is basically non-existent: it's just adding another attribute to the HTML field.

> The `btn-primary` class turns all the buttons dark-green, and takes the roundedness off them. Is that correct, as it doesn't look nice?

See #37 about the colour; I'll get onto that in a mo. It shouldn't remove the roundedness, no, but none of the buttons are rounded in any browser for me at the moment anyway.

> There is a fair amount of usability research that says that the "default" choice should be the obvious one, so making the other choice a link immediately relegates it.

This is what btn-primary does: it makes the primary action obviously different from a regular button (hence the name of the class). The change I'm suggesting is consistent with the research you allude to.

> Those come from the database, so are hard to style with the correct icon for the provider, or hyperlink (where would they hyperlink to?). Equally the buttons are submit buttons for a form. We could possibly replace it with an image for "remove".

The buttons I mentioned were the "associate your account with X" buttons, not any remove buttons. I _literally_ just get the text "Twitter", a heading below that, and three buttons for each of the social logins; no option to remove the association with the social accounts, or anything like that. It's basically useless for anything other than seeing that I have _an_ account from from the specified provider associated with my DDDEA account, and adding a new social account.

> Another issue with the form requiring a legend.

`form`s don't _have_ to have a fieldset, so if it's not appropriate to include a title (e.g., the Delete Session form), the fieldset (and so the legend) can be removed. Most of our forms have a title, though, usually in an `hN` tag before the form.

A further issue I've since noticed is that the @ character on the Twitter Handle field on the Edit Profile page is not vertically aligned with the text in the input field. They should share a common base-line.

adrianbanks:
The fact that you are not getting some display things suggest that you are missing some styles. Have you made sure that you have done a hard refresh? Don't know what's going on with the OAuth logins - it works well on my machine.

Regarding the `btn-primary` issue: the research suggests that the other actions, especially if they are non-destructive, shouldn't even be buttons (not using just colour to distinguish elements and all that WCAG/508 stuff) - hence the hyperlinks.

The rest are further improvements. 

To get this moving forwards (other smaller changes can't really happen with this pending without the risk of conflicts), **can we merge it**, check on the staging site that it works (as we currently have one machine where it does and one where it doesn't - pull it out if it doesn't work), then add the rest of these as separate improvements.? Otherwise, this pull request gets even bigger than it already is.

Further improvements:
- be able to dismiss the shouty box
- sort out the text on legends (sounds like we shouldn't be using fieldsets for most of the forms)
- use HTML5 input types where possible (bear in mind that while developers tend to run the latest versions of browsers, some people might be browsing from their work PC and not be using the latest-and-greatest)

Bug:
- the @ symbol doesn't line up with the text box on the profile page

Further discssions:
- styling of the shouty box
- `btn-primary` issue

philpursglove:
You've already seen how inept I am with UI/design stuff, so I hardly feel qualified to comment here. It's not going to stop me though... :-) 

On the Registered page, the 'update your profile' and 'submit a session' links aren't styled consistently. The colouring on the 'update your profile' link in particular is so subtle that to me it's difficult to tell it _is_ a link. 

In the Sessions view, something that I think might be beneficial to distinguish one session from another is shading of alternate rows on the page, maybe the same grey as on the sidebar?

Just thoughts, I don't see any of these as showstoppers...

adrianbanks:
This is how the two links looks on mine: 
screenshot

The links are style exactly the same according to Chrome's developer tools.

On the sessions page, I've added a separator between them (exactly the same as Amazon uses to separate search results). I get the impression Alastair is keen to not put background colours. One thing you may not have noticed it that the session are sorted alphabetically, but a user's own sessions will always appear at the top, with a different styling (similar to the sidebar) to make them stand out.

adrianbanks:
@alastairs Done the `btn-danger` change. Not sure what your other comment refers to (combining styles).

alastairs:
On 446b869 you have two separate style definitions for identical styling of related items. They could be rolled into one style definition.

PR #36, opened 2013-04-10
Add a database backup task

alastairs:
We probably should have had this before going live, but never mind... This pull request adds a couple of scripts to export the database as a BACPAC (a deployment artefact encompassing both schema and data) using the SqlPackage utility from SSDT. The resulting BACPAC can be right-click deployed to any SQL Server 2012 instance via SSMS.

I've added [an associated TeamCity build configuration](http://build.dddeastanglia.com/viewType.html?buildTypeId=bt7) which runs daily at 00.05. The configuration publishes the created BACPAC as an artefact, so we can download any backup from TeamCity.

I've deliberately sent this pull request to the production branch, because it has no effect on the product itself and so can safely skip the staging process. I have tested it against my own fork and watched the build fail and finally succeed. Given the nature of the task, I would like this to go live ASAP.


philpursglove:
I take it that's one of the things you got from Stacked?

----- Original Message -----

alastairs:
Indirectly, yes :smile: Was hoping for more on deploying databases, but I think EF's Migrations may take care of that for us. We need to test it out at some point soon...

philpursglove:
Saw the blob storage was set up when I logged onto Azure today, I downloaded the BACPAC later on and restored it into SQL2012 when I was looking at dsplaying the sessions...

----- Original Message -----

alastairs:
Yeah, the blob storage may be a useful thing to keep around: we can hook up a container to the Azure CDN and serve static files from it über-fast. Haven't decided on that one yet.

alastairs:
I'm assuming the otherwise deathly silence means this is good to go in...?

alastairs:
Actually, no it's really **NOT** ok to go in, just spotted a big flaw.

adrianbanks:
I'll trust your judgement on this as you know more about it (PS and SSDT) than I.

alastairs:
Man, I'm glad I caught that in time. I'd only committed the database connection string!

adrianbanks:
Have you expunged it from the history as well?

alastairs:
> Have you expunged it from the history as well?

As best I can. I re-wrote the history of the branch so it no longer appears in those commits, and I have expunged it from my local repository. According to GitHub the commits still exist on my fork and are accessible by their commitishes, but that can be resolved after this has been merged by deleting and re-forking the repository.

PR #45, opened 2013-04-11
Fix #37: Regenerate the Bootstrap stylesheet.

### LINKED ISSUE #37: Regenerate the Bootstrap stylesheet
The Bootstrap stylesheet needs regenerating: the `btn-primary` class, for example, currently displays quite horribly using DDDEA green as the background.

alastairs:
This is mostly to fix the .btn-primary rendering. I have also incorporated the hover colour into the Bootstrap stylesheet, so this is removed from the Site.css stylesheet.

I have documented the full set of overrides in #37 so we can refer back to it if we need to do this again.

### DIFF
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -9,8 +9,6 @@ dd {
 
 a { text-decoration: underline; }
 
-a:hover { color: #008600; }
-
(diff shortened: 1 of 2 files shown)


alastairs:
This is an opportunity to roll other styles into the bootstrap stylesheet/tweak things that don't quite work, so leave your requests here...

alastairs:
@adrianbanks Any comments before I hit merge on this?

Issue #41, opened 2013-04-10
Be able to dismiss the "shouty" box


adrianbanks:
Alastair is keen to be able to dismiss the shouty box that appears on some pages so that it doesn't get shown again.

Need to consider whether it reappears when the thing it is shouting about changes (e.g. session submission -> voting).

alastairs:
I just tried playing around with this with the Bootstrap alert functionality. It's not possible to permanently dismiss the banner because alerts are intended to be transitory: they appear once and you can dismiss it if you wish. 

To permanently dismiss the banner, we will need to set a cookie, that can be invalidated when the announcement is changed.

adrianbanks:
Fixed by #47.

PR #47, opened 2013-04-11
Issue #41: re-style the shouty box and allow it to be dismissed

adrianbanks:
The box now appears at the top of the page (like to old Stack Overflow notifications bar did). You can dismiss it and it will stay away, but only up until the next promoted event starts where it will reappear.

### DIFF
--- a/DDDEastAnglia/App_Start/BundleConfig.cs
+++ b/DDDEastAnglia/App_Start/BundleConfig.cs
@@ -8,7 +8,8 @@ namespace DDDEastAnglia
         public static void RegisterBundles(BundleCollection bundles)
         {
             bundles.Add(new ScriptBundle("~/bundles/jquery").Include(
-                        "~/Scripts/jquery-{version}.js", "~/Scripts/jquery.cookie.js"));
+                        "~/Scripts/jquery-{version}.js", 
+                        "~/Scripts/jquery.cookie.js"));
 
             bundles.Add(new ScriptBundle("~/bundles/jqueryui").Include(
                         "~/Scripts/jquery-ui-{version}.js"));
--- a/DDDEastAnglia/App_Start/BundleConfig.cs
+++ b/DDDEastAnglia/App_Start/BundleConfig.cs
@@ -8,7 +8,7 @@ namespace DDDEastAnglia
         public static void RegisterBundles(BundleCollection bundles)
         {
             bundles.Add(new ScriptBundle("~/bundles/jquery").Include(
-                        "~/Scripts/jquery-{version}.js"));
+                        "~/Scripts/jquery-{version}.js", "~/Scripts/jquery.cookie.js"));
 
             bundles.Add(new ScriptBundle("~/bundles/jqueryui").Include(
                         "~/Scripts/jquery-ui-{version}.js"));
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -236,11 +236,11 @@ fieldset {
 }
 
 #notifications-banner a:hover {
-    color: greenyellow;
+    color: #eee;
 }
 
 #notifications-banner .dismiss {
-    border: 2px solid white;
+    border: 2px solid #ccc;
     padding: 0 6px;
     cursor: pointer;
     float: right;
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -215,23 +215,27 @@ fieldset {
     padding-right: 10px;
 }
 
-.banner {
+#notifications-banner {
     color: white;
-    font-weight: bold;
-    font-size: 18px;
-    text-shadow: 0 1px 1px rgba(0, 0, 0, 0.75);
-    background-color: #006000;
-    background-image: linear-gradient(to bottom, #00b000, #006000);
-    border-radius: 10px;
-    padding: 15px;
-    margin: 30px 0 30px 0;
+    background-color: #008600;
+    text-shadow: 0 1px 1px black;
+    padding: 10px 20px 10px 20px;
+    margin: 0;
+    display: none;
+    border-bottom: 1px solid #777777;
 }
 
-.banner a {
-    color: white;
+#notifications-banner p {
+    margin: 0;
 }
 
-.banner a:hover {
+#notifications-banner a {
+    color: white;
+    text-decoration: none;
+    border-bottom: 1px solid white;
+}
+
+#notifications-banner a:hover {
     color: greenyellow;
 }
 
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -239,6 +239,21 @@ fieldset {
     color: greenyellow;
 }
 
+#notifications-banner .dismiss {
+    border: 2px solid white;
+    padding: 0 6px;
+    cursor: pointer;
+    float: right;
+}
+
+#notifications-banner .dismiss a,
+#notifications-banner .dismiss a:hover {
+    color: white;
+    font-weight: bold;
+    text-decoration: none;
+    border: 0;
+}
+
 a i {
     text-decoration: none;
 }
--- a/DDDEastAnglia/DataAccess/DataModel/Event.cs
+++ b/DDDEastAnglia/DataAccess/DataModel/Event.cs
@@ -64,8 +64,8 @@ namespace DDDEastAnglia.DataAccess.DataModel
         {
             PreConferenceAgendaItem votingStart;
             PreConferenceAgendaItem votingEnd;
-            if (_items.TryGetValue(DateType.SubmissionStarts, out votingStart)
-                && _items.TryGetValue(DateType.SubmissionEnds, out votingEnd))
+            if (_items.TryGetValue(DateType.VotingStarts, out votingStart)
+                && _items.TryGetValue(DateType.VotingEnds, out votingEnd))
             {
                 return votingStart.Date <= date
                        && date <= votingEnd.Date;
--- a/DDDEastAnglia/DataAccess/DataModel/Event.cs
+++ b/DDDEastAnglia/DataAccess/DataModel/Event.cs
@@ -64,8 +64,8 @@ namespace DDDEastAnglia.DataAccess.DataModel
         {
             PreConferenceAgendaItem votingStart;
             PreConferenceAgendaItem votingEnd;
-            if (_items.TryGetValue(DateType.VotingStarts, out votingStart)
-                && _items.TryGetValue(DateType.VotingEnds, out votingEnd))
+            if (_items.TryGetValue(DateType.SubmissionStarts, out votingStart)
+                && _items.TryGetValue(DateType.SubmissionEnds, out votingEnd))
             {
                 return votingStart.Date <= date
                        && date <= votingEnd.Date;
--- a/DDDEastAnglia/T4MVC.cs
+++ b/DDDEastAnglia/T4MVC.cs
@@ -100,6 +100,8 @@ namespace Links
         public static readonly string jquery_ui_1_10_2_js = T4MVCHelpers.IsProduction() && T4Extensions.FileExists(URLPATH + "/jquery-ui-1.10.2.min.js") ? Url("jquery-ui-1.10.2.min.js") : Url("jquery-ui-1.10.2.js");
                       
         public static readonly string jquery_ui_1_10_2_min_js = Url("jquery-ui-1.10.2.min.js");
+        public static readonly string jquery_cookie_js = T4MVCHelpers.IsProduction() && T4Extensions.FileExists(URLPATH + "/jquery.cookie.min.js") ? Url("jquery.cookie.min.js") : Url("jquery.cookie.js");
+                      
         public static readonly string jquery_unobtrusive_ajax_js = T4MVCHelpers.IsProduction() && T4Extensions.FileExists(URLPATH + "/jquery.unobtrusive-ajax.min.js") ? Url("jquery.unobtrusive-ajax.min.js") : Url("jquery.unobtrusive-ajax.js");
                       
         public static readonly string jquery_unobtrusive_ajax_min_js = Url("jquery.unobtrusive-ajax.min.js");
--- a/DDDEastAnglia/Views/Session/Index.cshtml
+++ b/DDDEastAnglia/Views/Session/Index.cshtml
@@ -10,7 +10,7 @@
 
 @if (new EventRepository().Get("DDDEA2013").CanSubmit())
 {
-    <p>Session submission is now open. If you would like to speak at <span class="dddea">DDD East Anglia</span>, please @Html.ActionLink("submit a session", "Create", "Session", null, new { @class = "submit-session" }).</p>
+    <p>Session submission is now open. If you would like to speak at @Html.DDDEastAnglia(), please @Html.ActionLink("submit a session", "Create", "Session", null, new { @class = "submit-session" }).</p>
 }
 
 @foreach (var session in Model)
--- a/DDDEastAnglia/Views/Session/Index.cshtml
+++ b/DDDEastAnglia/Views/Session/Index.cshtml
@@ -5,11 +5,6 @@
     ViewBag.Title = "Sessions";
 }
 
-@section featured
-{
-    @Html.Partial("_Banner")
-}
-
 <h2>Sessions</h2>
 
 @foreach (var session in Model)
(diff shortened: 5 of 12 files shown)


DDDEastAnglia/App_Start/BundleConfig.cs
```diff
@@ -8,7 +8,7 @@ public class BundleConfig
         public static void RegisterBundles(BundleCollection bundles)
         {
             bundles.Add(new ScriptBundle("~/bundles/jquery").Include(
-                        "~/Scripts/jquery-{version}.js"));
+                        "~/Scripts/jquery-{version}.js", "~/Scripts/jquery.cookie.js"));
```
alastairs: Please put the new script on a new line: it makes for easier diffing.

alastairs: Also, I wonder if we should bundle jQuery plugins separately. Don't really know enough about this new feature :disappointed:

DDDEastAnglia/Content/Site.css
```diff
@@ -215,26 +215,45 @@ fieldset {
...
     color: white;
-    font-weight: bold;
-    font-size: 18px;
-    text-shadow: 0 1px 1px rgba(0, 0, 0, 0.75);
-    background-color: #006000;
-    background-image: linear-gradient(to bottom, #00b000, #006000);
-    border-radius: 10px;
-    padding: 15px;
-    margin: 30px 0 30px 0;
+    background-color: #008600;
```
alastairs, line 220: Did you try this in the darker DDDEA green #013c01?

DDDEastAnglia/Content/Site.css
```diff
@@ -215,26 +215,45 @@ fieldset {
...
+    text-shadow: 0 1px 1px black;
+    padding: 10px 20px 10px 20px;
+    margin: 0;
+    display: none;
+    border-bottom: 1px solid #777777;
```
alastairs, line 225: Does it really need the bottom border? I would have thought the transition from green to white would be sufficient.

DDDEastAnglia/Content/Site.css
```diff
@@ -215,26 +215,45 @@ fieldset {
...
+
+#notifications-banner a {
     color: white;
+    text-decoration: none;
+    border-bottom: 1px solid white;
 }
 
-.banner a:hover {
+#notifications-banner a:hover {
     color: greenyellow;
```
alastairs: I don't much like this colour, and it's not used anywhere else in the site. Could we remove the underline on hover or something instead, or could we just have no hover effect? There's precedent for this in the carousel captions on the home page.

DDDEastAnglia/Content/Site.css
```diff
@@ -215,26 +215,45 @@ fieldset {
...
 }
 
+#notifications-banner .dismiss {
+    border: 2px solid white;
+    padding: 0 6px;
+    cursor: pointer;
```
alastairs, line 245: Surprised you need this. Why not set the `a` to `display: block` instead?

DDDEastAnglia/DataAccess/DataModel/Event.cs
```diff
@@ -64,8 +64,8 @@ public bool CanVote(DateTime date)
         {
             PreConferenceAgendaItem votingStart;
             PreConferenceAgendaItem votingEnd;
-            if (_items.TryGetValue(DateType.SubmissionStarts, out votingStart)
-                && _items.TryGetValue(DateType.SubmissionEnds, out votingEnd))
+            if (_items.TryGetValue(DateType.VotingStarts, out votingStart)
```
alastairs: Sorry if this is a bit of a pain to do, but is it possible to exclude this file from the change? I'm a bit antsy about these Entity Framework migrations and whether they're going to overwrite the database whenever anyone so much as sneezes near the data models.

DDDEastAnglia/T4MVC.cs
```diff
@@ -100,6 +100,8 @@ public static class Scripts {
         public static readonly string jquery_ui_1_10_2_js = T4MVCHelpers.IsProduction() && T4Extensions.FileExists(URLPATH + "/jquery-ui-1.10.2.min.js") ? Url("jquery-ui-1.10.2.min.js") : Url("jquery-ui-1.10.2.js");
                       
         public static readonly string jquery_ui_1_10_2_min_js = Url("jquery-ui-1.10.2.min.js");
+        public static readonly string jquery_cookie_js = T4MVCHelpers.IsProduction() && T4Extensions.FileExists(URLPATH + "/jquery.cookie.min.js") ? Url("jquery.cookie.min.js") : Url("jquery.cookie.js");
```
alastairs, line 103: This is unnecessary. In Release configuration, the Bundler will minify any JavaScript or CSS files passed through it, but not in Debug configuration. As such, you only need to include the non-minified version in the project and ensure it is included through the Bundler, an then you can debug your styles and scripts on your local machine, and have them minified for performance in production.

DDDEastAnglia/Views/Session/Index.cshtml
```diff
@@ -1,17 +1,18 @@
...
     ViewBag.Title = "Sessions";
 }
 
-@section featured
+<h2>Sessions</h2>
+
+@if (new EventRepository().Get("DDDEA2013").CanSubmit())
 {
-    @Html.Partial("_Banner")
+    <p>Session submission is now open. If you would like to speak at <span class="dddea">DDD East Anglia</span>, please @Html.ActionLink("submit a session", "Create", "Session", null, new { @class = "submit-session" }).</p>
```
alastairs: There's an HTML Helper for including the styled DDD East Anglia text, cunningly called `@Html.DDDEastAnglia()`.

DDDEastAnglia/App_Start/BundleConfig.cs
adrianbanks: I don't know much about it either. It also doesn't always appear to work (just been trying with something else and it was ignoring one of the includes).

DDDEastAnglia/Views/Shared/_Banner.cshtml
```diff
@@ -1,20 +1,25 @@
...
-    {
-        <div class="banner">
-            <p>Session voting is now open.</p>
-    
-            <p>Please @Html.ActionLink("vote on the sessions", "Vote", "Session") you would like to see at <span class="dddea">DDD East Anglia</span>.</p>
-        </div>                                
-    }
-</div>
+@{
+    var eventRepository = new EventRepository();
```
alastairs, line 5: Ouch, this is a lot of code to include in a View. Looks like a ViewModel is in order.

DDDEastAnglia/Views/Shared/_Banner.cshtml
```diff
@@ -1,20 +1,25 @@
...
+    var dddevent = eventRepository.Get("DDDEA2013");
+
+    var submissionEnds = dddevent.PreConferenceAgenda.First(i => i.DateType == DateType.SubmissionEnds);
+    var votingEnds = dddevent.PreConferenceAgenda.First(i => i.DateType == DateType.VotingEnds);
+}
+
+@if (@dddevent.CanSubmit())
+{
+    <div id="notifications-banner" expires-on="@submissionEnds.Date.ToString("R")">
+        <span class="dismiss"><a title="Dismiss this notification">X</a></span>
```
alastairs: Use the `&times;` HTML entity to get a nice square cross (ditto below).

DDDEastAnglia/Content/Site.css
adrianbanks, line 225: Yes, it just takes the harshness out of the transition. you can't really see the border, but you can see its effect.

DDDEastAnglia/Content/Site.css
adrianbanks, line 220: Yes. it is too dark. This green is the green of the navbar.

DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -3,7 +3,7 @@
 <head>
     <meta charset="utf-8" />
     <title>DDD East Anglia @ViewBag.Title</title>
-    <link href="http://www.dddeastanglia.com/favicon.ico")" rel="shortcut icon" type="image/x-icon" />
+    <link href="http://www.dddeastanglia.com/favicon.ico" rel="shortcut icon" type="image/x-icon" />
```
alastairs, line 6: Erk, where did that come from?? Good spot!

DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -140,7 +141,34 @@
         @Scripts.Render("~/bundles/bootstrap")
         @Styles.Render("~/Content/Markdown")
         @RenderSection("scripts", required: false)
+        
+        <script language="javascript">
```
alastairs, line 145: This would be good pulled out into a separate file so it can be passed through the bundler for minification.

DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -140,7 +141,34 @@
...
+            $(document).ready(function () {
+                if ($("#notifications-banner").length) {
```
alastairs: `if ($("#notifications-banner").length > 0)`? Damn JavaScript...

alastairs: Also, you could invert this to reduce nesting.

DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -140,7 +141,34 @@
...
+                    var eventChangeDateStr = $("#notifications-banner").attr("expires-on");
+                    var eventChangeDate = new Date(eventChangeDateStr);
 
+                    var setEventChangeDateStr = $.cookie("notification-dismissed");
+                    var setEventChangeDate = new Date(setEventChangeDateStr);
+
+                    if (setEventChangeDate < eventChangeDate) {
+                        @{/* if the date in the cookie is less than the date on the banner, the user dismissed an older banner */}
```
alastairs: Razor comments are `@* Your awesome comment *@`.

DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -140,7 +141,34 @@
         @Scripts.Render("~/bundles/bootstrap")
         @Styles.Render("~/Content/Markdown")
         @RenderSection("scripts", required: false)
+        
+        <script language="javascript">
+            $(document).ready(function () {
+                if ($("#notifications-banner").length) {
+                    var eventChangeDateStr = $("#notifications-banner").attr("expires-on");
```
alastairs: If you don't need these string variables elsewhere (i.e., they're temporary to create the Date variables), just inline them.

DDDEastAnglia/Views/Shared/_Banner.cshtml
```diff
@@ -1,20 +1,25 @@
...
+    var eventRepository = new EventRepository();
+    var dddevent = eventRepository.Get("DDDEA2013");
+
+    var submissionEnds = dddevent.PreConferenceAgenda.First(i => i.DateType == DateType.SubmissionEnds);
+    var votingEnds = dddevent.PreConferenceAgenda.First(i => i.DateType == DateType.VotingEnds);
+}
+
+@if (@dddevent.CanSubmit())
+{
+    <div id="notifications-banner" expires-on="@submissionEnds.Date.ToString("R")">
```
alastairs: `expires-on` should probably be called `data-expires` or similar, by convention.

DDDEastAnglia/Views/Shared/_Layout.cshtml
alastairs: Also see previous comment about the name of the `expires-on` attribute.

DDDEastAnglia/Content/Site.css
adrianbanks, line 245: Just tried that and it doesn't work. The `a` doesn't have a border applied, even when one is explicitly set.

DDDEastAnglia/Content/Site.css
alastairs, line 220: Ok, cool.

DDDEastAnglia/Content/Site.css
alastairs, line 225: Ok. (I haven't checked it out to eyeball it yet, just going by the code.)

DDDEastAnglia/Content/Site.css
adrianbanks: I've changed it to an ever-so slight grey hover colour, as I do think we need something to indicate a hover.

DDDEastAnglia/DataAccess/DataModel/Event.cs
adrianbanks: I'll take it out, but it will need to be fixed at some point before voting opens (the voting dates are actually using the submission dates). I'll add it as another issue.

adrianbanks: Issue #52.

DDDEastAnglia/T4MVC.cs
adrianbanks, line 103: I added the jquery plugin through NuGet. I didn't add this line of code - it got auto-generated by the T4MVC thingy.

DDDEastAnglia/Views/Session/Index.cshtml
adrianbanks: Ok. Seems a bit pointless (as the code is geared up to have many events, although we haven't written the views that way), but I've changed it

DDDEastAnglia/Views/Shared/_Banner.cshtml
adrianbanks, line 5: This banner gets included on every page in the site, rendered as a partial. Is it possible to use a viewmodel in this scenario? If so, how? That's why I did it this way.

DDDEastAnglia/Views/Shared/_Layout.cshtml
adrianbanks, line 145: I would do, but I'm having no end of trouble with the bundler. It keeps not including scripts that I have registered with it.

DDDEastAnglia/T4MVC.cs
alastairs, line 103: Ouch, ok.

DDDEastAnglia/Content/Site.css
alastairs, line 245: Oh. Did you change the HTML to match as well? I was thinking that if you set the `a` to `display: block`, you might not need to include the wrapping `.dismiss` element.

DDDEastAnglia/Views/Shared/_Layout.cshtml
adrianbanks: I think it is clearer nested as it is, especially with the weird condition check.

DDDEastAnglia/Content/Site.css
adrianbanks, line 245: The dismiss element is partly there for the jquery code. You need to either wrap the dismiss `a` or add a class it because it is not the only `a` in the banner. In short, I don't think it would be much simpler HTML and CSS if it did work.

DDDEastAnglia/Views/Shared/_Layout.cshtml
alastairs: Weird condition check? If you mean the .length check, you can make that much less weird using the suggestion above.

adrianbanks: Go on then, give me a clue. What would it look like?

alastairs: Exactly as a I wrote above: 

``` javascript
if ($("#notifications-banner").length > 0)
```

With the inversion to reduce nesting, it would be:

``` javascript
if ($("#notifications-banner").length <= 0) {
    return;
}
```

The code as-is is doing a "falsy" check, i.e., `0 == false` in JavaScript. If you want to keep it that way you could just do the following to invert it:

``` javascript
if (!$("#notifications-banner").length) {
    return;
}
```

DDDEastAnglia/Views/Shared/_Layout.cshtml
alastairs, line 145: Let's take a look at that tomorrow.

DDDEastAnglia/Views/Shared/_Layout.cshtml
adrianbanks: OIC. I was confused by your "suggestion above" comment - didn't know what you were referring to.

Issue #44, opened 2013-04-10
Add a link to this issue tracker in the footer of the site


adrianbanks:
As we have already had users reporting issues, let's encourage it.

philpursglove:
Agreed - Joel H-T tweeted me yesterday with his feature requests, I asked him to log them on the repo...

----- Original Message -----

alastairs:
I did a bit of a google around this topic earlier, and didn't come up with anything pre-made that we could use, sadly. However, perhaps with a bit of <del>plagiarism</del> inspiration from widgets like [UserVoice](https://www.uservoice.com/)'s and [Get Satisfaction](https://getsatisfaction.com/)'s, we could roll together a little JavaScript widget that posted back to the [GitHub REST API for Issues](http://developer.github.com/v3/issues/#create-an-issue).

adrianbanks:
I was thinking to keep it simple and just put a hyperlink to the issues page with the text something like "Found a problem?". No need for anything complicated (or for that matter time-consuming to code).

philpursglove:
I was thinking something similar...

----- Original Message -----

adrianbanks:
Fixed by #48. I've just added a link to the issues page.

PR #48, opened 2013-04-11
Issue #44: add a link to the github issues page in the footer

adrianbanks:


alastairs:
Something I noticed earlier: I meant for the dashes in the foot to be en or em dashes; the separation isn't great enough with a simple dash. Otherwise I think this is good to go.

Does `target="_blank"` open in a new tab?

adrianbanks:
Yes to the opening a new tab.

PR #50, opened 2013-04-11
Increase the size of the speaker icon on the session details

adrianbanks:
I've also increased the size of the speaker name, as it looked too small next to a larger icon.

### DIFF
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -194,6 +194,7 @@ fieldset {
 }
 
 .session-speaker {
+    font-size: 18px;
     margin-bottom: 10px;
 }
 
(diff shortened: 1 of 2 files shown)


DDDEastAnglia/Content/Site.css
```diff
@@ -194,6 +194,7 @@ fieldset {
 }
 
 .session-speaker {
```
alastairs: Shouldn't this be one of the heading classes instead (perhaps h3 or h4)? The text is, after all, semantically a heading.

DDDEastAnglia/Content/Site.css
```diff
@@ -187,8 +187,6 @@ fieldset {
 }
 
 .session-title {
```
alastairs: You can make this `session-details h3` now, and delete the `session-title` class (this is now captured through the semantic markup).

PR #58, opened 2013-04-13
Feature/voting

ninjaferret:
This should be some basic functionality for voting.

Added some voting controls to the Sessions Page, currently doing a form post-back.

Added a VoteController to manage all of the work

Added a database migration

Added Cookie Support

Removed votes are determined by **IsVote = 0** in the database so the total votes for a session is the count of votes for the session where **IsVote = 1** minus the total where the **IsVote = 0**... now I'm thinking I should have just made it an int and used 1 and -1... that's just another migration...

I wanted to get the basic functions in. It's not all unit tested yet but the controller at least is working. Still to do:

Javascript voting - we can fall back on the form mechanism if JS is not enabled
Add in more of the "extra information" fields to the vote (e.g. user Agent etc)


ninjaferret:
Closing as my other branch has a complete voting solution

PR #60, opened 2013-04-14
Fix #9: Feature/enhanced voting

### LINKED ISSUE #9: We need to have voting enabled
The idea at the moment is to use cookies to store the user's preferences

I think we want a simple table that stores as much info on the user as we can. IP Address, Session ID, Cookie ID, Location, browser, etc

ninjaferret:
This should be voting done to a greater/lesser extent. It is progressively enhanced so the buttons will do a postback to the vote controller if Javascript is not enabled and the Javascript will take control over submission

The information that we store is:
- SessionId (maybe we should have called them talks... )
- CookieID (a guid associated with the cookies used to store the user's votes)
- Time Recorded
- WebSessionId
- UserId (if they are logged in)
- IPAddress
- User Agent
- Referrer
- Screen Resolution (at the moment just from the javascript)

The controller is mostly tested, the return values aren't tested but what is saved to the database is.


DDDEastAnglia/Content/Site.css
```diff
@@ -260,3 +260,21 @@ a i {
 .speaker {
     border-bottom: 1px solid lightgray;
 }
+
+.voting {
+    float: right;
+}
+
+.voting button.remove-vote {
```
alastairs: Could we stick to the default Bootstrap buttons for consistency across the site? There are additional classes to adjust the style (e.g., colour) of the button if desired.

DDDEastAnglia/Controllers/VoteController.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        private readonly IVotingCookieRepository _votingCookieRepository;
+        private readonly IVoteRepository _voteRepository;
+        private readonly ISessionRepository _sessionRepository;
+        private readonly IEventRepository _eventRepository;
+        private readonly ITimeProvider _timeProvider;
+        private readonly IRequestInformationProvider _requestInformationProvider;
+
+        public VoteController() 
+            : this(new VotingCookieRepository(),
```
alastairs: Wow. Is there a Facade that could be introduced over some of these dependencies? I try to keep my class dependencies down to two as much as possible, with no more than three.

DDDEastAnglia/Controllers/VoteController.cs
```diff
@@ -0,0 +1,152 @@
...
+            var currentEvent = _eventRepository.Get("DDDEA2013");
+            var result = new SessionVoteModel();
+            result.SessionId = sessionId;
+            result.UserCanVote = currentEvent != null && currentEvent.CanVote();
+            result.VotedForByUser = cookie.Contains(sessionId);
+            return PartialView(result);
+        }
+    }
+
+    public class VoteModel
```
alastairs: Should be in a different file.

DDDEastAnglia/DataAccess/EntityFramework/DDDEAContext.cs
```diff
@@ -12,5 +13,6 @@ public DDDEAContext()
 
         public DbSet<UserProfile> UserProfiles { get; set; }
         public DbSet<Session> Sessions { get; set; }
+        public DbSet<Vote> Vote { get; set; }
```
alastairs: Should be `Votes` for consistency.

DDDEastAnglia/DataAccess/EntityFramework/Migrations/201303302150305_InitialCreate.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace DDDEastAnglia.DataAccess.EntityFramework.Migrations
+{
+    public partial class InitialCreate : DbMigration
+    {
+        public override void Up()
+        {
+            CreateTable(
+                "dbo.UserProfile",
+                c => new
```
alastairs: What is `c`?

DDDEastAnglia/Models/VotingCookie.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace DDDEastAnglia.Models
+{
+    public class VotingCookie
+    {
+        public const string CookieName = "DDDEA2013.Voting";
+        public static readonly DateTime DefaultExpiry = new DateTime(2013, 4, 30);
+        private readonly List<int> _sessionsVotedFor = new List<int>();
+
+        public VotingCookie(string name)
```
alastairs: Do we really need all these constructors? How many of them are in use?

DDDEastAnglia/Scripts/on-page-voting.js
```diff
@@ -0,0 +1,29 @@
+﻿
```
alastairs: Why "on page" voting? Why not simply "voting"?

Also, I don't see any tests for the JavaScript code here.

DDDEastAnglia/Views/Vote/Status.cshtml
```diff
@@ -0,0 +1,21 @@
+﻿@model DDDEastAnglia.Models.SessionVoteModel
+
+@if (Model.UserCanVote && Model.VotedForByUser)
```
alastairs: Could the `UserCanVote` check be moved elsewhere? It's a pain to have to check this in both conditions, which could otherwise be a simple `if..else` (rather than the more complicated `if..else if`).

DDDEastAnglia/Views/Vote/Status.cshtml
```diff
@@ -0,0 +1,21 @@
...
+{
+    <span id="@("Vote" + Model.SessionId)" class="voting">
```
alastairs: This span is identical in the other block, so could be moved outside of both.

alastairs:
A couple of nitpicky things: you have misspelled "Referrer" as "Referer" all over the tests, there are commented-out tests (worse than commented-out code!), and there are a couple of random whitespace things too (extra line breaks, missing line breaks).

DDDEastAnglia/Controllers/VoteController.cs
ninjaferret: I don't understand, really this needs to be merged back in with the status method again

DDDEastAnglia/Views/Vote/Status.cshtml
ninjaferret: So you are happy for an empty voting span when the voting isn't open?

DDDEastAnglia/Controllers/VoteController.cs
ninjaferret: I agree, however, I wanted to get this in rather than cleaning this up right now. For example the VoteRepository and the Cookie Repository could be merged into a single VoteRepository with a HttpVoteRepository that takes another IVoteRepository passes the calls through but manages the cookie.

DDDEastAnglia/DataAccess/EntityFramework/Migrations/201303302150305_InitialCreate.cs
ninjaferret: This is created by the Migrations code, it was either created by default when I turned on migrations, or it was already there, it's not something I paid much attention to

DDDEastAnglia/Views/Vote/Status.cshtml
alastairs: Yeah, I think so. The only issue is the call to `Model.SessionId` if there are no sessions submitted, but that should be taken care of by the controller anyway...

DDDEastAnglia/Views/Vote/Status.cshtml
alastairs: Perhaps this could be moved to the controller.

ninjaferret: I'm adding a view for "Voting not allowed" which just returns nothing

DDDEastAnglia/Scripts/on-page-voting.js
ninjaferret: OK renamed it

DDDEastAnglia/Controllers/VoteController.cs
alastairs: Ok.

DDDEastAnglia/Controllers/VoteController.cs
alastairs: All I'm saying is this model class should be in a separate file, in the Models namespace.

DDDEastAnglia/Views/Vote/Status.cshtml
alastairs: It should do something other than returning nothing, e.g., displaying a message stating that voting is not open/has closed.

alastairs:
One further thing, after discussion with @adrianbanks. Could you ensure the database migration is atomic for this change, i.e., it starts at the point the schema is at now, and after a single migration step ends up at the schema as you need it? The interim migration steps (particularly the `DROP Votes` one!) are a bit scary.

adrianbanks:
Is there something I have to do to get the database working? I have errors saying that I have a missing Conferences table. Should the db migration have happened automatically, or do I need to do something manually?

ninjaferret:
No idea, it all worked for me as I applied the migrations, sometimes I had to cause IIS Express to re-load the site to make the migrations apply - what tables do you have? Has it applied any of the tables? If not, that is usually an indication it hasn't tried to do any of the migrations

To answer Alastair's question, what is dangerous about the migrations? Firstly, no votes are being process so we have no data to screw up, secondly where does it DROP Votes? It drops some columns from the votes table but my dataset has retained data throughout these changes. I have no idea how to make the migrations atomic for this change at the moment

ninjaferret:
Personally, having used them for a bit, I find the whole migrations thing a bit distasteful. Code first migrations actually relies on auto-generated code that also relies on the current state of the developer's database and the code that they are writing. 

You cannot write them manually because it auto-generates some kind hash, or encrypted string, of the changes that it thinks are needed; this does also make it difficult to significantly amend the generated code because it still thinks that there are changes outstanding and you need to add a new migration. 

Given it's dependency on the actual database you are connected to it probably also means that it doesn't play nicely with two developers both making changes to the models independently. Each one will be potentially starting from a different point and the migrations may not work because if you pull in a new migration but the source has changed I'm not sure what is going to happen so we will have to be a bit careful about that. 

I may be able to make a single atomic database, maybe if I undo all the migrations on it and try to start from scratch and re-apply every modification that I made, I'll have to get back to you on that one.

alastairs:
> I may be able to make a single atomic database, maybe if I undo all the migrations on it and try to start from scratch and re-apply every modification that I made, I'll have to get back to you on that one.

If that's what it's going to take, don't bother. Sounds like a move away from EF is in order too?

alastairs:
I'm in the process of merging this manually. There were a few conflicts, most of which have been easy to resolve. 

The changes you've made to (i.e. the introduction of) the `BannerController` are incompatible with @adrianbanks's introduction of a `BannerController` which has already made its way into `staging`. As a result, I've thrown away your `BannerController` in favour of Adrian's, and so there may be work here that you want/need to re-do. 

I've also not really done much of a review of your latest changes, as this pull request is just too big now (134 files changed?!), and so it's difficult to find what I've already reviewed and what has been added since. It would help if the change were more focussed on the voting feature.

ninjaferret:
It doesn't have to be merged if there are too many issues with the EF
migrations and the size of the work, I can drop that branch and do it again.

The changes that you suggested were already started with a different piece
of work that I was doing so I decided to finish that but if it's too big I
can start again.

On 21 April 2013 18:47, Alastair Smith notifications@github.com wrote:

> I'm in the process of merging this manually. There were a few conflicts,
> most of which have been easy to resolve.
> 
> The changes you've made to (i.e. the introduction of) the BannerControllerare incompatible with
> @adrianbanks https://github.com/adrianbanks's introduction of a
> BannerController which has already made its way into staging. As a
> result, there may be work here that you want/need to re-do.
> 
> I've also not really done much of a review of your latest changes, as this
> pull request is just too big now (134 files changed?!), and so it's
> difficult to find what I've already reviewed and what has been added since.
> It would help if the change were more focussed on the voting feature.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/dddeastanglia/DDDEastAnglia/pull/60#issuecomment-16738098
> .

## 

Ian Johnson

alastairs:
I don't think starting again is an option given that we're opening voting in a little over a week, and it's not fair on you when you've put so much time into this already.

PR #62, opened 2013-04-14
Fix #41: re-style the shouty box and allow it to be dismissed

### LINKED ISSUE #41: Be able to dismiss the "shouty" box

adrianbanks:
This is a redo of https://github.com/dddeastanglia/DDDEastAnglia/pull/47 that got lost from the staging branch when Alastair removed some merges.


alastairs:
I will merge this manually from @adrianbanks's branch to production as it should be there already.

Issue #34, opened 2013-04-10
Session details page does not link to Speaker Bio


JoelHT-Landmark:
The session details page (e.g. http://www.dddeastanglia.com/Session/Details/17) has a header giving the speaker's name and avatar, but doesn't provide a link to the speaker bio.

See also Issue #33

adrianbanks:
This has now been done as part of https://github.com/dddeastanglia/DDDEastAnglia/pull/69.

PR #69, opened 2013-04-16
Add a speaker details page and link to it from the sessions

adrianbanks:
This pull request adds a speaker page, so that the details of an individual speaker can be shown/linked to. This page is linked to from each session.

To do this, I've extracted a partial view for showing the speaker details, and refactored some code in the controllers.

I've also styled the speakers page to be consistent with the sessions page so that a user's own profile will be highlighted and have a link to edit it.

### DIFF
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -208,11 +208,7 @@ fieldset {
     display: none;
 }
 
-.actions {
-    float: right;    
-}
-
-.actions .action {
+.action {
     padding-right: 10px;
 }
 
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -176,7 +176,7 @@ fieldset {
     margin-bottom: 15px;
 }
 
-.session-details-own {
+.session-details-own, .speaker-own {
     border: 1px solid #e3e3e3;
     border-radius: 5px;
     background-color: #f5f5f5;
@@ -261,4 +261,5 @@ a i {
 
 .speaker {
     border-bottom: 1px solid lightgray;
+    padding: 10px;
 }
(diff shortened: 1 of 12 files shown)


adrianbanks:
I've just noticed that this might address issue #34. Didn't mean it to, just thought it would be useful to link to the speaker.

DDDEastAnglia/Content/Site.css
```diff
@@ -208,11 +208,11 @@ fieldset {
     display: none;
 }
 
-.session-actions {
-    text-align: right;    
+.actions {
```
alastairs: No need for this class, you can use `pull-right` from Bootstrap.

DDDEastAnglia/Views/Speaker/Details.cshtml
```diff
@@ -0,0 +1,7 @@
+﻿@model DDDEastAnglia.Models.SpeakerDisplayModel
+
+@{
+    ViewBag.Title = "Details";
```
alastairs, line 4: Title should be set on the model really, not hard-coded in the view.

alastairs:
Would benefit from some tests once Ian's change has gone in.

DDDEastAnglia/Views/Speaker/Details.cshtml
adrianbanks, line 4: None of the other views currently do this, so I'd suggest that we target them all as a separate issue.

DDDEastAnglia/Content/Site.css
adrianbanks: Done.

DDDEastAnglia/Views/Speaker/Details.cshtml
alastairs, line 4: Ok.

PR #78, opened 2013-04-17
Fix #74: Style error messages from forms consistently

### LINKED ISSUE #74: Style error messages from forms consistently
They should use the bootstrap styles that a few messages currently use.

adrianbanks:
This change style all of the forms using bootstrap's horizontal form styles.

It also puts the validation summary at the top of the form, and styles it using bootstrap.

### DIFF
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -132,8 +132,23 @@ a.btn, .dropdown-menu a, a.carousel-control { text-decoration: none; }
 }
 
 .field-validation-error, .validation-summary-errors {
-    color: red;
+    color: #b94a48;
     font-weight: bold;
+}
+
+.validation-summary-errors {
+    background-color: #f2dede;
+    border-color: #eed3d7;    
+    padding: 8px 14px;
+    margin-bottom: 20px;
+    text-shadow: 0 1px 0 rgba(255, 255, 255, 0.5);
+    -webkit-border-radius: 4px;
+    -moz-border-radius: 4px;
+    border-radius: 4px;
+}
+
+.validation-summary-errors ul {
+    margin: 0;
 }
 
 .validation-summary-errors li {
--- a/DDDEastAnglia/Views/Account/ChangePassword.cshtml
+++ b/DDDEastAnglia/Views/Account/ChangePassword.cshtml
@@ -13,35 +13,42 @@ else
         <div class="alert alert-success">@ViewBag.Message</div>
     }
     
-    using (Html.BeginForm("ChangePassword", "Account")) 
+    using (Html.BeginForm("ChangePassword", "Account", FormMethod.Post, new { @class = "form-horizontal"})) 
     {
         @Html.AntiForgeryToken()
 
         <fieldset>
             <legend>Please enter the following details</legend>
-            <ol>
-                <li>
-                    @Html.LabelFor(m => m.OldPassword)
+
+            @Html.ValidationSummary(true)
+
+            <div class="control-group">
+                @Html.LabelFor(m => m.OldPassword, new { @class = "control-label"})
+                <div class="controls">
                     @Html.PasswordFor(m => m.OldPassword)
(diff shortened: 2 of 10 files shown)


DDDEastAnglia/Views/Account/ChangePassword.cshtml
```diff
@@ -13,35 +13,42 @@ else
...
-                </li>
-            </ol>
-        
-            @Html.ValidationSummary(true)
+                </div>
+            </div>
         
-            <p>
+            <div>
                 <input type="submit" class="btn" value="Change password" />
```
alastairs, line 50: `btn-primary`. I thought you'd caught all these in another pull request? There are other forms in this pull request missing this class too.

DDDEastAnglia/Content/Site.css
```diff
@@ -132,10 +132,25 @@ a.btn, .dropdown-menu a, a.carousel-control { text-decoration: none; }
 }
 
 .field-validation-error, .validation-summary-errors {
```
alastairs, line 134: This looks like you've duplicated the Bootstrap style information into our own file rather than re-using the defined Bootstrap classes?

DDDEastAnglia/Views/Account/ChangePassword.cshtml
adrianbanks, line 50: I have. It hasn't been merged yet, so I'm not changing them on this branch also to avoid conflicts.

DDDEastAnglia/Content/Site.css
adrianbanks, line 134: I have yes. I have no (real) control over the html markup that the validation outputs. I could either have duplicated the css styles, or created a custom helper to output the validation summary which involved writing a class that emitted lots of html. I chose the former as it is easier to implement and to update (and also means we don't accidentally drop back into using the standard validation summary).

DDDEastAnglia/Views/Account/ChangePassword.cshtml
adrianbanks, line 50: I am mistaken. It has been merged, but it is not in production yet.

DDDEastAnglia/Content/Site.css
alastairs, line 134: There's no hook into MVC to change/alter the emitted classes? I had thought they added support for this in v3 or v4.

alastairs, line 134: Yeah, does [this](http://msdn.microsoft.com/en-us/library/dd470919.aspx) help at all? [ValidationSummaryValidCssClassName](http://msdn.microsoft.com/en-us/library/system.web.mvc.htmlhelper.validationsummaryvalidcssclassname.aspx), etc.

adrianbanks, line 134: I couldn't find anything when I searched, and StackOverflow suggested that this wasn't possible.

I'll take another look tonight.

adrianbanks, line 134: Can't do anything about this. Those fields of `HtmlHelper` are read only, so cannot be changed. There doesn't appear to be any other way of doing it.

alastairs, line 134: Yuck. [Reflection hack](http://stackoverflow.com/a/6026517)?

adrianbanks, line 134: No, I'd prefer it the way it is. I have a plan to introduce sass, so the duplication of the styles can be solved later using that.

alastairs, line 134: Ok, that may be a better way to solve it. 

Could we perhaps use Less instead? I know it's not quite as nice as Sass, but I don't think there's much in it, and Bootstrap uses Less.  We could then do our customisations to Bootstrap from the source rather than having to set all the customisations on the website and download a custom build each time.

adrianbanks, line 134: Yes, could do. I'll take a closer look at each when I make the change.

PR #79, opened 2013-04-19
Added list of speakers at the head of the page with individual links

### LINKED ISSUE #33: Speakers page does not include in-page anchor tags to allow direct navigation to a speaker
On the speakers page (http://www.dddeastanglia.com/Speaker), each speaker's entry is listed, but there is no way to navigate directly to the section of the page for a given speaker.

This page should include an in-page anchor tag so that the following URL would navigate directly to my entry (for example)

http://www.dddeastanglia.com/Speaker#Joel%20Hammond-Turner

Additionally, this page could show a "limited" bio (say 4 lines), and link to a profile page specific to the speaker - this would allow speakers to write long bios in their profile without the main speaker's page becoming ugly.

For example 

http://www.dddeastanglia.com/Speaker/12

would link to the full-page profile.

philpursglove:
Fixes #33.

### DIFF
--- a/DDDEastAnglia/Views/Speaker/Index.cshtml
+++ b/DDDEastAnglia/Views/Speaker/Index.cshtml
@@ -7,9 +7,17 @@
 
 <h2>Speakers</h2>
 
+<ul>
+@foreach (SpeakerDisplayModel speaker in Model)
+{
+    <li><a href="#@speaker.Username">@speaker.Name</a></li>
+}   
+</ul>
+
 @foreach (var speaker in Model)
 {
     <div class="row-fluid">
+        <a id="@speaker.Username" />
         @Html.Partial("_SpeakerDetailsPartial", speaker)
     </div>
 }
--- a/DDDEastAnglia/Views/Speaker/Index.cshtml
+++ b/DDDEastAnglia/Views/Speaker/Index.cshtml
@@ -7,17 +7,9 @@
 
 <h2>Speakers</h2>
 
-<ul>
-@foreach (SpeakerDisplayModel speaker in Model)
-{
-    <li><a href="#@speaker.Username">@speaker.Name</a></li>
-}   
-</ul>
-
 @foreach (var speaker in Model)
 {
     <div class="row-fluid">
-        <a id="@speaker.Username" />
         @Html.Partial("_SpeakerDetailsPartial", speaker)
     </div>
 }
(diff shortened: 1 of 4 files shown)


philpursglove:
Don't know that I'm _hugely_ happy with the list at the top of the page, but I don't know that I can come up with another way to surface the links.

DDDEastAnglia/Views/Speaker/Index.cshtml
```diff
@@ -7,9 +7,17 @@
...
+@foreach (SpeakerDisplayModel speaker in Model)
+{
+    <li><a href="#@speaker.Username">@speaker.Name</a></li>
+}   
+</ul>
+
 @foreach (var speaker in Model)
 {
     <div class="row-fluid">
+        <a id="@speaker.Username" />
```
adrianbanks: This id can go into the partial view (`_SpeakerDetailsPartial`), and instead of adding it as a `a` tag, the id can be put onto the `h3` tag that is already there.

DDDEastAnglia/Views/Speaker/Index.cshtml
```diff
@@ -7,9 +7,17 @@
 
 <h2>Speakers</h2>
 
+<ul>
+@foreach (SpeakerDisplayModel speaker in Model)
+{
+    <li><a href="#@speaker.Username">@speaker.Name</a></li>
+}   
+</ul>
```
adrianbanks: I had a chat with Alastair about this yesterday. Now that there is a detailed speaker page (i.e. each speaker has their own page that shows just their details - http://dddeastanglia.com/Speaker/Details/23), there is less impetus to have an anchor to link directly to the speaker's part of the whole list. It is still valid to put the anchors in just in case, but there is probably no need to have the list at the top of the main speakers list.

DDDEastAnglia/Views/Speaker/Index.cshtml
philpursglove: Didn't know you could put that on a <h> tag and it still work if you put into the URL e.g. \Speakers#philpursglove, the only times I've ever done this was using an <a> tag

DDDEastAnglia/Views/Speaker/Index.cshtml
philpursglove: Indeed, hence my comment above about the list being there...

alastairs: Ok, let's take out the list. @adrianbanks and I came up with an alternative solution, which we've documented in #81.

PR #83, opened 2013-04-19
Sort the speakers the same as the sessions

adrianbanks:
This sorts them alphabetically by name, with the current speaker first.

### DIFF
--- a/DDDEastAnglia/Controllers/SpeakerController.cs
+++ b/DDDEastAnglia/Controllers/SpeakerController.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Web.Mvc;
 using DDDEastAnglia.DataAccess;
@@ -27,6 +28,7 @@ namespace DDDEastAnglia.Controllers
                 }
             }
 
+            speakers.Sort(new SpeakerDisplayModelComparer());
             return View(speakers);
         }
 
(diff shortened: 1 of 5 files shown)


DDDEastAnglia/Controllers/SpeakerController.cs
```diff
@@ -27,6 +28,7 @@ public virtual ActionResult Index()
                 }
             }
 
+            speakers.Sort(new SpeakerDisplayModelComparer());
```
alastairs, line 30: Out of curiosity, is this any better or worse than using `OrderBy()`?

DDDEastAnglia/Controllers/SpeakerController.cs
```diff
@@ -66,4 +68,22 @@ private SpeakerDisplayModel CreateDisplayModel(UserProfile userProfile, IEnumera
                 };
         }
     }
+
+    public class SpeakerDisplayModelComparer : IComparer<SpeakerDisplayModel>
```
alastairs: Should be in a separate file. Or, as it's an implementation detail, a private nested class.

DDDEastAnglia/Controllers/SpeakerController.cs
adrianbanks, line 30: Yes, you can't do custom (complex) sorting using `OrderBy()`. I want the current user's profile (and sessions in the other controller) to be at the top.

alastairs, line 30: Hmm, I thought you could, but perhaps that's only in Linq2Sql/Entity Framework.

PR #87, opened 2013-04-20
Add automapper

philpursglove:
Bringing in AutoMapper to do some of the model->viewmodel transformations

### DIFF
--- a/DDDEastAnglia/Controllers/SessionController.cs
+++ b/DDDEastAnglia/Controllers/SessionController.cs
@@ -154,12 +154,18 @@ namespace DDDEastAnglia.Controllers
             var isUsersSession = Request.IsAuthenticated && session.SpeakerUserName == User.Identity.Name;
             var tweetLink = CreateTweetLink(isUsersSession, session.Title, Url.Action("Details", "Session", new { id = session.SessionId }, Request.Url.Scheme));
 
-            SessionDisplayModel displayModel = Mapper.Map<Session, SessionDisplayModel>(session);
-            displayModel.SpeakerId = profile.UserId;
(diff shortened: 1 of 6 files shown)


DDDEastAnglia/Controllers/SessionController.cs
```diff
@@ -149,26 +152,21 @@ public virtual ActionResult DeleteConfirmed(int id)
         private SessionDisplayModel CreateDisplayModel(Session session, UserProfile profile)
         {
             var isUsersSession = Request.IsAuthenticated && session.SpeakerUserName == User.Identity.Name;
-            var tweetLink = CreateTweetLink(isUsersSession, session.Title, Url.Action("Details", "Session", new {id = session.SessionId}, Request.Url.Scheme));
+            var tweetLink = CreateTweetLink(isUsersSession, session.Title, Url.Action("Details", "Session", new { id = session.SessionId }, Request.Url.Scheme));
```
alastairs, line 155: I wonder how much AutoMapper is buying us here: you've only saved a couple of lines of code with it.

philpursglove, line 155: Yeah, I'll revert this out (or can I remove commit b73ae02 from the pull request?) - my thought process was along the lines of 'If I'm bringing in Automapper, I must justify bringing it in by using it as much as possible'

alastairs:
I think this might be a bit premature in general, actually. It will certainly become more useful as we build out more functionality, but I think there's re-architecting to go in before this buys us enough to warrant its inclusion.

I'll hold off closing it without merge for the moment if you think there's a good reason to include it now.

alastairs:
I've not heard any objections, so closing now.

PR #97, opened 2013-04-21
Remove excess padding from the bottom of a speaker

adrianbanks:
This had the effect of having an odd block of whitespace at the top of each speaker (except the first) when in a list, especially when coupled with the horizontal rule that we use as a separator.

### DIFF
--- a/DDDEastAnglia/Content/Site.css
+++ b/DDDEastAnglia/Content/Site.css
@@ -125,10 +125,6 @@ a.btn, .dropdown-menu a, a.carousel-control { text-decoration: none; }
     margin-bottom: 10px;
 }
 
-.speaker {
-    margin-bottom: 30px;
-}
-
 .inline-large-icon {
     display: inline-block;
     width: 20px;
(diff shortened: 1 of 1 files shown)


alastairs:
How much padding-bottom is rendered now?

adrianbanks:
None in addition to the standard padding applied by bootstrap to the row-fluid class. It looks much better balanced now, as there is equal space both above and below each speaker.

alastairs:
> It looks much better balanced now, as there is equal space both above and below each speaker.

I agree it looks more balanced now, and that 30px is a bit large with the horizontal rule in place, but I think the sessions look a bit cramped and squashed together without _any_ extra whitespace between them. I had a bit of a play around, and changing the style definition to `margin: 10px 0` looks quite nice, giving them a bit more room to breathe.

adrianbanks:
`margin-top: 10px` works ok, I'll go with that. I don't think we should apply margins everywhere - just where we want the extra space.

alastairs:
The same change should also be applied to the Sessions' page, for consistency. I notice that there are two `.speaker` definitions in the file, too: the other is at the bottom of the file.

adrianbanks:
I'll tweak my pull request to do that...

adrianbanks:
Done.

PR #99, opened 2013-04-21
Fix #81: add a search box to the sessions and speaker pages

### LINKED ISSUE #81: Quickly find a session/speaker within the lists on the relevant pages
At the top of the sessions and speakers page, add a jquery filter/search to find a particular session/speaker.

adrianbanks:


DDDEastAnglia/Scripts/searchfilter.js
```diff
@@ -0,0 +1,12 @@
+﻿function filter(containerClass, element) {
+    var value = $(element).val();
```
alastairs: What are `element` and `value`?

DDDEastAnglia/Views/Shared/_SearchBox.cshtml
```diff
@@ -0,0 +1,3 @@
+﻿<div class="pull-right form-search" style="margin-bottom:30px">
```
adrianbanks: I'll pull this style into the stylesheet.

DDDEastAnglia/Views/Shared/_SearchBox.cshtml
```diff
@@ -0,0 +1,3 @@
+﻿<div class="pull-right form-search">
+    <input class="search-query" type="text" placeholder="Search..." onkeyup="filter('@ViewData["containerClass"]', this)" />
```
alastairs: `ViewData`: probably good enough for now, but should perhaps be replaced with something more appropriate (= strongly-typed) in future.

(`ViewData` is the precursor to `ViewBag`, which was introduced in MVC 3 when .NET 4 introduced dynamic typing, so in general it should be avoided even more than `ViewBag`!)

adrianbanks: Possibly, but I thought it way over the top to add a new model (and potentially a new controller given some of the examples I was looking at) for this.

alastairs:
:+1: Doing the manual merge now.

alastairs:
Thought I'd take a quick look before pushing to staging, and I get this error: "Section already defined: "scripts"."

```
Source Error: 

Line 18: }
Line 19: 
Line 20: @foreach (var session in Model.Sessions)
Line 21: {
Line 22:     Html.RenderPartial("_SessionDetailsPartial", session);

Source File: z:\Code\DDDEastAnglia\DDDEastAnglia\Views\Session\Index.cshtml    Line: 20 
```

adrianbanks:
Where do you get that error?

alastairs:
On the Sessions page.

adrianbanks:
Are you looking at my branch, or a merge of it? I don't get that error (on my branch).

If you are looking at it when merged to staging, I think it's a bad merge. Ian has introduced a scripts block on the session index view, and so have I. They will need to be merged together.

alastairs:
> If you are looking at it when merged to staging, I think it's a bad merge. Ian has introduced a scripts block on the session index view, and so have I. They will need to be merged together.

Ok, will resolve that manually. I've had no end of problems tonight with git automatically undoing changes that people have made. 

I have managed to view this on the Speakers page, and the search box needs to be pushed down a bit relative to the heading text. I haven't played around with it, so can't give you an accurate figure, but I would say it's 5~10px that it needs to shuffle down.

It might also be nice to put a magnifying glass icon on the search box too.

I don't think it's correct to use an ellipsis (or a simulated ellipsis) in the placeholder text. Convention is that it is used for things that will open a new dialog (e.g., on a menu item), and I don't remember seeing it used anywhere on the web.

adrianbanks:
I'll add some padding at the top.

I did add a magnifying glass icon, but the bootstrap styles dont work together when the text box is rounded. It also makes the search box too prominent in the page, and makes it appear that you need to submit once you've typed, which you don't. I can't add it as a watermark, as the placeholders don't support HTML.

As for the trailing ellipsis, I've seen some places that do use them (chrome uses it in its find in page search box, for example). I'll try it without and see if it looks ok.

alastairs:
> I did add a magnifying glass icon, but the bootstrap styles dont work together when the text box is rounded.

Try using a `<button />` (or perhaps an `<a />`) instead of a `<span />` as the `input-append`, as described in the [Search Form](http://twitter.github.io/bootstrap/base-css.html#forms) section of the documentation.

Also, [this Stack Overflow question](http://stackoverflow.com/questions/9550738/bootstrap-navbar-search-icon) has some useful tips. Note we don't currently have in our project the glyphicons sprite mentioned in the first answer, as we're using Font Awesome for icons instead. 

> chrome uses it in its find in page search box, for example

Chrome is desktop software, though, not a website. Google, Bing, Trello, Apple and Amazon don't use placeholder text at all; eBuyer does use placeholder text, but doesn't use an ellipsis. Google doesn't use any adornments of its search box because it's synonymous with search; all the others use a magnifying glass or the word "Search" (the latter is only used by eBuyer, who also include the word "Search" in their placeholder text).

adrianbanks:
I've added some padding and removed the ellipsis.

I haven't added an icon because I don't think it needs one as it makes the search box too prominent with the default bootstrap styling, and also it seems impossible to add one using the bootstrap styling without completely redefining the bootstrap styles, sort of defeating the point of using the bootstrap.

alastairs:
> also it seems impossible to add one using the bootstrap styling without completely redefining the bootstrap styles

Did you have a read of that Stack Overflow thread? That indicates it is possible.

adrianbanks:
Yes. They have redefined the styles to make it work (i.e. not used bootstrap exclusively).

alastairs:
We've overridden the default styles elsewhere, and you can, if you prefer, take the approach that the Bootstrap documentation takes of adding an extra class for customisations:

``` html
<input type="text" class="search-query search-icon" />
```

...and the CSS:

``` css
.search-query.search-icon {
    /* Add necessary styles here */
}
```

adrianbanks:
I don't think it needs an icon, especially if it requires overriding lots of styling.

alastairs:
The point I'm making is that, by convention (your argument for most things UI and UX-related), search boxes _do_ have an icon, and [this example](http://jsfiddle.net/hYAEQ/) achieves that with just two extra styles.

adrianbanks:
> by convention (your argument for most things UI and UX-related), search boxes do have an icon

That is not a true convention. Some search boxes do have an icon, some don't. 

I'm trying hard to come up with an example where searching within a page (like the search boxes I've added) has an icon, but struggling, especially when the user doesn't have to submit the search (i.e. the search happens as you type). All of the examples I've found for this kind of instant search don't have icons.

Adding the icon, especially so when using the bootstrap append behaviour, makes the icon look like a submit button for the search. It also makes the search box too prominent in the page.

alastairs:
Ok, I'll get this merged now (including fixing the dodgy auto-resolution).

alastairs:
Just tried it out - can you make the filtering case-insensitive? It's ready to go in once that's done.

adrianbanks:
@alastairs Happy for this to be merged now?

alastairs:
:+1: Doing the manual merge now.

alastairs:
Pulling out this pull request: it is the source of the deployment errors. Please rebase onto staging, ensuring the below files are removed from the project, and resubmit.

These files were re-added in the pull request because your new searchfilter.js file was added amongst these files, which were present on your branch but removed on the staging branch. Git interpreted the addition of searchfilter.js as adding the entire hunk, which brought back these files (or at least their references in the project file) from the dead.

Files in error:
- Scripts\tweetbutton.js
- ProfileController.generated.cs
- SessionController.generated.cs
- SharedController.generated.cs
- SpeakerController.generated.cs
- T4MVC.cs

alastairs:
On the plus side, your other pull request can be merged now if you can reopen it fairly easily.

adrianbanks:
I'll rebase this tonight.

PR #100, opened 2013-04-21
Fix #53: The tweet links are (sometimes) including the session url twice.

### LINKED ISSUE #53: The tweet links are (sometimes) including the session url twice.
I've had it do this intermittently. Haven't pinned it down yet as what the cause is. Sometimes the link is there twice, other times not. I _suspect_ that it is the Twitter js doing something, as when it appears twice, the tweet is selected up to the end of the first url, but the second url is not selected.

alastairs:
Introduced a new HTML Helper for the Tweet button (see 3aec308 for documentation of the escaping and encoding needed). Removed, thoroughly, the old Tweet button.

### DIFF
--- a/DDDEastAnglia/App_Code/HTMLExtensions.cs
+++ b/DDDEastAnglia/App_Code/HTMLExtensions.cs
@@ -52,4 +52,20 @@ public static class HTMLExtensions
         }
         return "";
     }
+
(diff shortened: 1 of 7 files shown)


adrianbanks:
:+1: Looks good.

PR #102, opened 2013-04-22
EntityFramework keeping cached state for voting

ninjaferret:
This will, for the moment, fix the issue with the EntityFramework caching state preventing any underlying database changes (e.g. changing the StartDate for CalendarItems by directly going to the database)

Going forward I think we will want to find a more elegant solution, along the lines of UnitOfWork but I just want to get this change in and work on that separately


alastairs:
Looks good :+1: 

> Going forward I think we will want to find a more elegant solution, along the lines of UnitOfWork but I just want to get this change in and work on that separately

Yes. I have a couple of re-architectings in mind to improve some items in our code base.

Issue #96, opened 2013-04-21
Admin Area is b0rked.


alastairs:
I tried visiting http://dddeastaging.azurewebsites.net/admin/adminhome as a newly-registered user, and got the standard unhelpful YSOD (Runtime Error, an error occurred, set customErrors to off, etc.).

If the user is not in the Admin role, a 404 Not Found response should be returned.

alastairs:
This YSOD is also on production.

adrianbanks:
I don't think that this is (currently) down to not being logged in as an admin. It seems to always throw an exception ("The controller for path '/admin/adminhome' was not found or does not implement IController.").

alastairs:
This appears to be a routing issue as far as I can tell. `AdminHomeController` does implement `IController`, and so it must be that it can't be found.

ninjaferret:
I think I know why, moving the banner to a controller action in the
_Layout.cshtml ties us to the standard controllers.

When rendering the Admin area it can't find the controller because it can't
look outside it's own namespace?

Maybe a simple solution is to change the default layout to something
simpler for the administrator pages?

On 22 April 2013 14:21, Alastair Smith notifications@github.com wrote:

> This appears to be a routing issue as far as I can tell.
> AdminHomeController does implement IController, and so it must be that it
> can't be found.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/dddeastanglia/DDDEastAnglia/issues/96#issuecomment-16785239
> .

## 

Ian Johnson

adrianbanks:
Yes, that might make it easier. Currently, most of the links in the layout are broken due to them being in the admin folder. Removing them (and having a "back to main site" link?) would solve this as well.

ninjaferret:
I think for any urls/actions etc we would probably specify the area in the
route value objects

Html.ActionLink("Display Text", "Action", "Banner", new { area = "" }, null
);

However, for the admin stuff we don't need a lot of the general bumpfh that
the proles need, like the sponsors and the timeline etc, might be better to
have a set of admin related headers and as you say a main "back to site"
link

On 22 April 2013 15:12, Adrian Banks notifications@github.com wrote:

> Yes, that might make it easier. Currently, most of the links in the layout
> are broken due to them being in the admin folder. Removing them (and having
> a "back to main site" link?) would solve this as well.
> 
> —
[...]

PR #103, opened 2013-04-22
Fix the T4MVC errors I introduced.

alastairs:
I was doing merges yesterday on my laptop, which doesn't play nicely with the T4MVC bits at all. This commit resolves the issues I inadvertantly introduced, and all the issues @adrianbanks raised yesterday _except_ the admin area issue (#96).


alastairs:
"errors" == "inconsistencies" or similar. I don't think it caused any actual problems, just some weirdness with the generated files.

alastairs:
Merging as no one has screamed yet.

PR #107, opened 2013-04-22
Remove T4MVC

alastairs:
T4MVC is actively preventing me from working on the site, so I've decided to pull it out. It doesn't seem to be buying us enough to be worth the pain its causing.


alastairs:
@adrianbanks @ninjaferret Can this go in?

adrianbanks:
It's hard to see from the diffs if it has broken anything, so merge it and we'll check on staging.

alastairs:
Merged. Please give the staging site a good hammering: all the links and all the forms need testing. :unamused:

adrianbanks:
I've clicked on everything I can think of and it all seems fine, with the exception of a stray `;` in the header causing an unwanted spacing: 



Not sure where it came from yet, so may not be this change that has caused it.

The admin page still crashes, but I think that is not caused by this.

PR #108, opened 2013-04-22
Fix #96: Add a new Layout for the Admin Area.

### LINKED ISSUE #96: Admin Area is b0rked.

alastairs:
I've included a start of a new layout for the Admin Area, but it will need more work in future.

This fixes the immediate issue, but there's a problem with the Create Session link in the admin area: its using the Admin Route and so 404s. We can at least access the admin pages again.

**Note:** This builds on #107, so **do not merge** until that has gone in.

### DIFF
--- a/DDDEastAnglia/App_Start/BundleConfig.cs
+++ b/DDDEastAnglia/App_Start/BundleConfig.cs
@@ -36,7 +36,11 @@ namespace DDDEastAnglia
             AddScriptBundle(bundles, "~/bundles/voting", "~/Scripts/voting.js");
 
             AddStyleBundle(bundles, "~/Content/admin",
-                           "~/Content/admin.css");
+                           "~/Content/bootstrap.css",
+                           "~/Areas/Admin/Content/admin.css",
+                           "~/Content/media-queries.css",
+                           "~/Content/font-awesome.css",
+                           "~/Content/font-awesome-ie7.min.css");
 
             AddStyleBundle(bundles, "~/Content/css",
                            "~/Content/bootstrap.css",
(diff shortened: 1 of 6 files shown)


alastairs:
I'm just rebasing this to make it easier to merge and verify.

alastairs:
Although GitHub is saying it can't auto-merge this, there are no manual conflicts to resolve: all are auto-resolved by git.

alastairs:
Is this ok to merge @adrianbanks @ninjaferret?

adrianbanks:
It looks ok. One question though: why is the "Conference Admin" a link that goes nowhere?

alastairs:
It's a temporary URL. I'll replace it with the proper URL in a mo.

alastairs:
This is good to merge now, diffs verified.

adrianbanks:
The log out link on the drop down menu has lost its style (position and hover).

alastairs:
Good catch.

adrianbanks:
Happy to merge as is though, as I've got a branch based off this and I'd like to rebase it.

alastairs:
Shouldn't be too tricky to fix. Will take a quick look and update if I can (will probably be a bit of a hacky fix).

alastairs:
Fixed with a copy-paste job. See f0d7190 for details of a proper fix; I didn't put that in because it would have been a larger change and require a bit of trial-and-error to work out what in Site.css is site-specific and what is a generally-applicable Bootstrap override.

PR #110, opened 2013-04-23
Make it a little harder for people to vote for their own sessions

ninjaferret:
If a speaker is logged in we know which session is theirs or not so don't display the voting button if the session belongs to them

### DIFF
--- a/DDDEastAnglia/Views/Session/_SessionDetailsPartial.cshtml
+++ b/DDDEastAnglia/Views/Session/_SessionDetailsPartial.cshtml
@@ -12,9 +12,13 @@
             </span>
         </div>
     }
+    else
+    {
+        Html.Action("Status", "Vote", new {id = @Model.SessionId});
+    }
     <h3>
         @Model.SessionTitle
-        @Html.Action("Status", "Vote", new { id = @Model.SessionId })
+        
     </h3>
     <h4>
         <img src="@Model.SpeakerGravitarUrl" width="32" height="32" alt="@Model.SpeakerName">


alastairs:
I guess this prevents them from voting for their own session twice, but they can still vote for other people's sessions twice. Non-speakers who have registered can vote twice.

Doesn't this put speakers at a disadvantage, albeit a small one?

ninjaferret:
The code behind prevents them from voting for the same session twice, well,
unless they remove the cookie that I set. They can always log-out and vote
for their session ;-) (as the cookie would stay the same we could easily
find the people who do that)

Anyway, it has the side-effect of removing a bit of ugly UI where the
button and the "actions" are butted up too close together

On 23 April 2013 10:23, Alastair Smith notifications@github.com wrote:

> I guess this prevents them from voting for their own session twice, but
> they can still vote for other people's sessions twice. Doesn't this put
> them at a disadvantage, albeit a small one?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/dddeastanglia/DDDEastAnglia/pull/110#issuecomment-16847784
> .

## 

Ian Johnson

alastairs:
> They can always log-out and vote for their session

Sorry, that's what I meant by voting for a session twice. Anyone can "legitimately" vote twice, once by being logged out and once by being logged in, the exception being speakers who can't vote for _some sessions_ when they're logged in. 

Seems like the fairest/most equal way to do it is to allow everyone to vote twice in this manner, on any session.

ninjaferret:
The only way they can vote twice for something is to remove the cookie that
we set, but that means they can vote as many times as they like. Logging
out would only make it so that people could vote for their own session.

Maybe you're right though, just means that I have to find another way to
make the page look decent

On 23 April 2013 11:02, Alastair Smith notifications@github.com wrote:

> They can always log-out and vote for their session
> 
> Sorry, that's what I meant by voting for a session twice. Anyone can
> "legitimately" vote twice, once by being logged out and once by being
> logged in, the exception being speakers who can't vote for _some sessions_when they're logged in.
> 
> Seems like the fairest/most equal way to do it is to allow everyone to
> vote twice in this manner, on any session.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/dddeastanglia/DDDEastAnglia/pull/110#issuecomment-16849404
> .

## 

Ian Johnson

ninjaferret:
OK, decided that we will let people vote for their own sessions

PR #124, opened 2013-04-25
Fix #123: Only need one @script tag in the Session/Index.cshtml

### LINKED ISSUE #123: Sessions List is broken on staging
Just returns an error page. Might be related to the fix @ninjaferret put in earlier?

ninjaferret:
Maybe another merge issue but the index.cshtml ended up with two @script sections


alastairs:
Oh, yes. I resolved that in a previous merge of the search feature, but completely forgot about it last night.

PR #129, opened 2013-04-25
Fix #86: add the ability for a user to reset their password

### LINKED ISSUE #86: There is no ability to reset a forgotten password
I'm not sure whether this should be supported out of the box by the membership stuff, but there doesn't appear to be any way of doing it on the site.

adrianbanks:
This uses SendGrid on Azure to send emails.

:warning: **To do this, there needs to be an SMTP server configured. The details are in the appSettings file in the dropbox account.** You will need the latest version of this file to make this work locally.


alastairs:
Tests?

alastairs:
Fixes #86.

adrianbanks:
I did look at writing tests, but couldn't figure out what to test? One class is a controller, one sends email - both hard to unit test.

alastairs:
> I did look at writing tests, but couldn't figure out what to test? One class is a controller, one sends email - both hard to unit test.

Controllers are not hard to test. If your controllers _are_ hard to test, they need refactoring to be more testable, because all you've done is write legacy code. 

In the controller you can test things like the `ActionResult` returned (use MvcContrib.TestHelper to make this easier), whether the appropriate calls were made to the services it depends on, etc. 

Your email service class can be tested in a similar way. It's all about putting in the correct abstractions and making appropriate use of dependency injection and other similar techniques.

adrianbanks:
If I abstract the bits that can't be tested out of the email class, I'm not sure there is anything left to test other than a `string.Replace()` (which even then doesn't _have_ to actually replace anything), which is why I didn't go that far when writing it.

To test the controller, it would make most sense to introduce a DI framework (e.g. Ninject), so that the relevant dependencies can be constructor-injected instead of just `new`ed inside the methods. That way, I can assert that the controller is coordinating the correct classes. As it stands, I can't do this.

alastairs:
Looks like there's a fair bit to test: ensuring the details of the message (sender, from address, etc.) are correct, ensuring the body is correct, ensuring the appropriate API method is called. 

You can use, as a temporary solution to the Constructor Injection problem, the Bastard Injection anti-pattern. We can then get an IoC container in as a separate change.

alastairs:
I've added the necessary configuration settings to the production and staging sites on Azure, so this should "just work" when deployed.

adrianbanks:
Working on testing the email stuff. We could really do with a DI framework after this change, but I'll refrain from that in this change.

adrianbanks:
I've reshaped the code and introduced some tests. I can't test the controller using MvcContrib unless I recompile it a check it in, as there doesn't appear to be an MVC4 version available on NuGet.

alastairs:
Assembly binding redirects in the web.config will make that work fine. You can check GiveCampUK/GiveCRM for an example, IIRC.

adrianbanks:
Those bindings are already present in the web.config of the main app, but the tests using `TestHelper` still fail.

> MvcContrib.TestHelper.ActionResultAssertionException : Expected result to be of type ViewResult. It is actually of type ViewResult.

Googling for this suggests that the fix is to recompile the MvcContrib library against the latest version of MVC. This would mean that we have to check this recompiled version into git. :disappointed:

[FluentMVCTesting](http://teststack.github.io/pages/fluentmvctesting.html) looks good, and does appear to work correctly. However, the controller isn't easy to test, as it just creates a new `DDDEAContext`, meaning I cannot inject a test version. This would be much easier once Ninject is in.

alastairs:
> However, the controller isn't easy to test, as it just creates a new DDDEAContext, meaning I cannot inject a test version.

Can you not use Bastard Injection again as a temporary fix?

alastairs:
> Googling for this suggests that the fix is to recompile the MvcContrib library against the latest version of MVC. 

That's weird, I've definitely solved that before using assembly binding redirects, at least twice (GiveCRM and an orchestra website I'm working on).

Anyway, if FluentMVCTesting is nicer and works OOTB, then we should go for that instead.

adrianbanks:
> Can you not use Bastard Injection again as a temporary fix?

Not really, because I don't create the controller - MVC does. The only way I can think is to have two constructors, one parameterless that calls the other one that is used for testing. Even then, I need to manipulate the `ModelState` object which I can't seem to do in tests.

> That's weird, I've definitely solved that before using assembly binding redirects, at least twice (GiveCRM and an orchestra website I'm working on).

The discussions I've seen suggest that it works in VS2010, but not in VS2012. FluentMVCTesting seems nicer out of the box, but its syntax isn't as nice.

DDDEastAnglia.Tests/Helpers/Email/Smtp/SmtpHostSettings_Should.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        [Test]
+        public void UseTheStandardSmtpPort_WhenNoPortIsSupplied()
+        {
+            var hostSettings = new SmtpHostSettings("host");
+            Assert.That(hostSettings.Port, Is.EqualTo(SmtpHostSettings.DefaultSmtpPort));
+        }
+
+        [Test]
+        public void AllowNoUsernameToBeSet()
```
alastairs, line 24: Is this right? There aren't many open relays around these days are there?

DDDEastAnglia/ForgottenPasswordTemplate.html
```diff
@@ -0,0 +1,27 @@
+﻿<html>
```
alastairs, line 1: Have you checked this in a mail reader (other than Gmail), e.g. Outlook?

DDDEastAnglia.Tests/Helpers/Email/Smtp/SmtpHostSettings_Should.cs
adrianbanks, line 24: If we are designing this application to be deployable for other organisers to use, then we should allow this. If you are hosting the site on a machine that has an SMTP server installed (e.g. IIS with SMTP added), it would be allowable to not have to authenticate to connect to it if it only accepted connections from localhost (or used something like IPSec).

DDDEastAnglia/ForgottenPasswordTemplate.html
adrianbanks, line 1: It is sort of "borrowed" from the Evernote password reset email, so it should work (or theirs doesn't).

DDDEastAnglia/Helpers/AppSettings/IAppSettingsProvider.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace DDDEastAnglia.Helpers.AppSettings
+{
+    public interface IAppSettingsProvider
```
alastairs, line 3: Interesting. I don't normally go to these lengths (I don't think it buys much for the extra levels of indirection introduced); instead I normally just hook up my strongly-typed configuration classes directly to the ConfigurationManager.

DDDEastAnglia/Helpers/Email/ResetPasswordEmailSender.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Net.Mail;
+using DDDEastAnglia.Helpers.File;
+
+namespace DDDEastAnglia.Helpers.Email
+{
+    public class ResetPasswordEmailSender
+    {
+        public const string FromEmailAddress = "admin@dddeastanglia.com";
```
alastairs, line 14: Why public?

adrianbanks, line 14: I have a test that checks that the address the reset being sent from has been set correctly. This is the address that it checks against. Should ultimately come from the site config tables in the database once we have them.

DDDEastAnglia/Helpers/AppSettings/IAppSettingsProvider.cs
adrianbanks, line 3: To do that means that the `SmtpHostSettingsProvider` class would then be tied to `ConfigurationManager`. This abstraction makes the tests slightly easier to manage.

DDDEastAnglia/Helpers/Email/Smtp/SmtpHostSettingsProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            host = appSettingsProvider.GetSetting(SmtpHostKey);
+            port = int.Parse(appSettingsProvider.GetSetting(SmtpPortKey));
+            username = appSettingsProvider.GetSetting(SmtpUsernameKey);
+            password = appSettingsProvider.GetSetting(SmtpPasswordKey);
+        }
+
+        public IMailHostSettings GetSettings()
+        {
+            return new SmtpHostSettings(host, port, username, password);
```
alastairs: As you're not keeping the created instance of `SmtpHostSettings` around, this is a factory rather than a provider.

DDDEastAnglia/Helpers/Email/SendGrid/SendGridEmailSender.cs
```diff
@@ -0,0 +1,31 @@
...
+            this.hostSettings = hostSettings;
+        }
+
+        public void Send(IMailMessage message)
+        {
+            var credentials = new NetworkCredential(hostSettings.Username, hostSettings.Password);
+            SMTP instance = SMTP.GetInstance(credentials, hostSettings.Host, hostSettings.Port);
+            
+            // don't like this, but will do for now
+            var sendGridMessageWrapper = (SendGridMessageWrapper) message;
```
alastairs, line 27: Interesting that you wrapped up the SendGridMessage, but not the Singleton which, to me at least, is the nastier bit of the API.

DDDEastAnglia/Helpers/File/FileContentsProvider.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace DDDEastAnglia.Helpers.File
+{
+    public class FileContentsProvider : IFileContentsProvider
+    {
+        public string GetFileContents(string path)
+        {
+            using (var reader = new StreamReader(path))
+            {
+                string contents = reader.ReadToEnd();
+                return contents;
```
alastairs, line 12: `return reader.ReadToEnd();` ? Or does that not work well because the `StreamReader` is disposable?

DDDEastAnglia/Models/ResetPasswordStepOneModel.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace DDDEastAnglia.Models
+{
+    public class ResetPasswordStepOneModel
```
alastairs, line 5: You decided to stick with separate fields in the end, then?

DDDEastAnglia/Helpers/Email/Smtp/SmtpHostSettingsProvider.cs
adrianbanks: I was keeping it at one point, but changed it so that it would pick up the config settings live without needing an app restart. I'll rename the class.

DDDEastAnglia/Helpers/File/FileContentsProvider.cs
adrianbanks, line 12: This makes it easier to debug. Otherwise, you cannot see what it has read, especially if the return value of the method is not assigned to a variable.

DDDEastAnglia/Models/ResetPasswordStepThreeModel.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace DDDEastAnglia.Models
+{
+    public class ResetPasswordStepThreeModel
+    {
+        [Required]
+        public string ResetToken { get; set; }
+
+        [Required]
+        [StringLength(100, ErrorMessage = "The new password must be at least {2} characters long.", MinimumLength = 6)]
```
alastairs: Was it this model that the validation was failing? Have you definitely got the format string index right?

DDDEastAnglia/Helpers/Email/SendGrid/SendGridEmailSender.cs
adrianbanks, line 27: I didn't think it worth doing as there was nothing of value to be tested without using the real service to test against.

DDDEastAnglia/Views/ResetPassword/Step1.cshtml
```diff
@@ -0,0 +1,36 @@
+﻿@model DDDEastAnglia.Models.ResetPasswordStepOneModel
+
+<h2>Reset your password</h2>
+
+@using (Html.BeginForm("ResetPassword", "ResetPassword", FormMethod.Post, new {@class = "form-horizontal"}))
```
alastairs: Spaces around the braces, whilst you're making updates :smile:

DDDEastAnglia/Views/ResetPassword/Step3.cshtml
```diff
@@ -0,0 +1,36 @@
+﻿@model DDDEastAnglia.Models.ResetPasswordStepThreeModel
+
+<h2>Reset your password</h2>
+
+@using (Html.BeginForm("SaveNewPassword", "ResetPassword", FormMethod.Post, new {@class = "form-horizontal"}))
```
alastairs: Spaces here too.

DDDEastAnglia/Models/ResetPasswordStepThreeModel.cs
adrianbanks: That doesn't look correct. I'll check.

DDDEastAnglia/Models/ResetPasswordStepOneModel.cs
adrianbanks, line 5: We have two registered users whose usernames are an email address. As this email address can technically be different from the user's defined email address, I kept the two fields.

alastairs:
> The only way I can think is to have two constructors, one parameterless that calls the other one that is used for testing.

That's exactly what the Bastard Injection anti-pattern is, and even though it's an anti-pattern, it will get us out of this testing hole we're currently in without having to a) introduce an IoC container or b) take the alternative proper approach to DI in MVC, which is creating our own Controller Factory (which will be thrown away when an IoC container comes in).

adrianbanks:
How do I manipulate the `ModelState` and also check that it has been manipulated in the tests?

DDDEastAnglia/Models/ResetPasswordStepThreeModel.cs
adrianbanks: Yes this is correct: 
- `{0}` is the name of the field
- `{1}` is the maximum length
- `{2}` is the minimum length

DDDEastAnglia/ForgottenPasswordTemplate.html
adrianbanks, line 1: Just checked it in Outlook (web) and it looks ok.

DDDEastAnglia/Models/ResetPasswordStepThreeModel.cs
alastairs: Ok, and "100" is the maximum length?

alastairs:
> How do I manipulate the `ModelState` and also check that it has been manipulated in the tests?

I don't think you do: that would be a bit like testing the framework, wouldn't it? Instead, you can call your controller action(s) with a model you know to be invalid and check the behaviour is as you expect.

alastairs:
How are you getting on with this pull request?

DDDEastAnglia/Models/ResetPasswordStepThreeModel.cs
adrianbanks: Yes, although I will probably change the attribute to a `MinLength` one instead so that there it no upper limit on the length.

adrianbanks:
Haven't had chance to test the controller yet.

adrianbanks:
@alastairs I've got lost in the comments here to know which ones are still relevant. Is there anything outstanding on this, or can it be merged?

alastairs:
I've left a couple of comments on the tests you've added. Beyond that, it's still missing tests on the Controller(s), but it's basically there now.

adrianbanks:
Added tests for the controller.

adrianbanks:
@alastairs Is this good to merge now?

alastairs:
:+1:

PR #130, opened 2013-04-25
Fix #72: Remove (most) legends and fieldsets as they were not appropriate.

### LINKED ISSUE #72: Address the use of legends on the forms
The legends on lots of the form aren't proper legends.

adrianbanks:
I've left a few in, because they do work in those cases.

### DIFF
--- a/DDDEastAnglia/Areas/Admin/Views/Session/Delete.cshtml
+++ b/DDDEastAnglia/Areas/Admin/Views/Session/Delete.cshtml
@@ -7,38 +7,37 @@
 <h2>Delete</h2>
 
 <h3>Are you sure you want to delete this?</h3>
-<fieldset>
-    <legend>Session</legend>
 
-    <div class="display-label">
-         @Html.DisplayNameFor(model => model.Title)
-    </div>
-    <div class="display-field">
-        @Html.DisplayFor(model => model.Title)
-    </div>
+<div class="display-label">
+        @Html.DisplayNameFor(model => model.Title)
+</div>
+<div class="display-field">
+    @Html.DisplayFor(model => model.Title)
+</div>
 
-    <div class="display-label">
-         @Html.DisplayNameFor(model => model.Abstract)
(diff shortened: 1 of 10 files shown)


DDDEastAnglia/Areas/Admin/Views/Session/Delete.cshtml
```diff
@@ -7,38 +7,37 @@
 <h2>Delete</h2>
 
 <h3>Are you sure you want to delete this?</h3>
-<fieldset>
```
alastairs, line 10: We had a fieldset outside of a form? That's invalid HTML5 anyway.

alastairs:
I really don't think the correct fix is to remove the fieldset and legend from the majority of the forms. The fieldset element groups related form controls, and the legend element provides a title for that group. I think the more correct fix is to move the text from the heading elements into the legend element.

This is important from an accessibility perspective, as correctly using fieldset and legend elements vastly improves the experience for people using screen readers.

adrianbanks:
> I think the more correct fix is to move the text from the heading elements into the legend element.

That would not be correct either. The `fieldset` element groups sets of controls. The spec says that a `fieldset` must also have a `legend`. The legend should describe what the fieldset represents, much like the `alt` tag on an image (i.e. a description of the image, not what it represents). The heading text of almost all of the pages doesn't fit this role.

I've just followed these semantic rules on several of the pages I've removed the fieldsets from to see how it works. The page either looks too cluttered (due to the bootstrap styling of legends, or the fact that there is seeming pointless text within the forms), contains text that duplicates a field label (in the case where there is only one logical input in a group), or doesn't really have any logical grouping of input controls (e.g the delete session page).

The only page that the legend makes sense on is the account registration page, and possibly the edit profile page.

The fieldsets and legends make it semantically better/correct, but I'm not sure it works well otherwise, especially when you read about their support in the mainstream screen readers, which seem to handle it badly.

PR #132, opened 2013-04-26
Fix #116: rename gravitar to gravatar

### LINKED ISSUE #116: GravatarUrl is spelled incorrectly in the UserProfile model
It is spelled "GravitarUrl".

adrianbanks:


alastairs:
This is failing the build.

adrianbanks:
This is down to staging being ahead of production. I branched from production, but there are changes in staging that don't build with this change integrated into them.

I've reabsed this branch onto staging to fix this, but it demonstrates that production needs to move forwards more frequently.

alastairs:
Yes it does, but we agreed to hold off moving production forward for now. A more appropriate resolution would have been to have left the branch off production and merge in the branches you need, as we do at work, but ok.

PR #137, opened 2013-04-29
Fix #133: Add the extra parameter to make the links work

### LINKED ISSUE #133: The links to edit/delete a session don't work
They are in the form `Session/Edit?Length=7` and `Session/Delete?Length=7` which is clearly wrong as they should contain the session id. When you click on them, you get a 404 error.

ninjaferret:


alastairs:
Please could you use a named parameter to make it obvious what this extra parameter is?

alastairs:
Merging manually due to conflicts. Out of interest, why does this fix the problem?

ninjaferret:
FYI It was simply that the ActionLink overload on the page was:

**ActionLink**(string _linkText_, string _actionName_, object _routeValues_, object _htmlAttributes_)

So the _ActionLink("Edit", "Edit", "Session", new { id= Model.Id})_ mapped the string "Session" to the route values and apparently if you pass a string in as route values you get the **length=7** thing. Adding the extra parameter told it to use this overload:

**ActionLink**(string _linkText_, string _actionName_, string _controllerName_, object _routeValues_, object _htmlAttributes_)

I guess I could have removed the controller name to achieve the same thing but this seemed more appropriate given we were deliberately specifying the controller

alastairs:
Ewwwwwwwwwwwwwwwwwwwwww..............

(Thanks for clarifying.)

PR #138, opened 2013-04-29
Database Project

alastairs:
:+1: Good to merge :+1: 

**Tasks**
- [x] Fix the build hang
- [x] Update with latest schema
- [x] Add `__MigrationHistory` table data


alastairs:
Woop woop! The process hang is fixed. 

For interested parties (@adrianbanks, perhaps?), the fix is to set the MSBUILDDISABLENODEREUSE environment variable to 1: MSBuild nodes are kept around by default after the job has been completed for performance purposes.

Now, to fix up the schema...

alastairs:
Ok, that updates the schema. I haven't yet managed to work out how to add the data to the `__MigrationHistory` table, though, so we're not out of the migration/deployment woods yet...

alastairs:
The `__MigrationHistory` table data is added via a Post-Deployment script: essentially, the quickest, dirtiest way I could work out how to do this. There doesn't seem to be a whole lot of documentation about using EF Code-First Migrations with SSDT yet, probably because of that whole "Code-First" thing.

As mentioned in the commit message on 739d2e8, there's a high probability that using this approach will mean ensuring this script is updated with all future migrations, which is an unsustainable approach. This might be another reason to ditch EF in favour of another ORM.

alastairs:
More searching seems to imply that Entity Framework doesn't play nicely with Microsoft's own tooling (even the Model-First approach doesn't work properly with SSDT database projects).

As such, unless there are any objections, I think we should merge this as-is, taking the initial pain of keeping the post-deployment script up-to-date with migration details, and rip out EF sometime in the not-too-distant (but possibly post-DDDEA) future.

alastairs:
Haven't heard any objections, so merging now.

PR #143, opened 2013-04-30
Fix #142: Typo in the default end date for the voting cookie

### LINKED ISSUE #142: Voting Buttons are b0rked
Clicking the Vote button against any of the sessions on the staging site (or a development site) seems to have no effect at all. Anyone else seeing this?

ninjaferret:
Fix the default end-date for the voting cookies - we will need to make come from the database at some point


alastairs:
Good catch, that would have been unfortunate.

PR #144, opened 2013-04-30
Fix #136

### LINKED ISSUE #136: Randomize order of sessions during voting
To prevent the first sessions getting all the votes.

SimonPStevens:
Randomly order the sessions when voting is active.


alastairs:
Can you get some tests around the change you made to the SessionController? It would be good to get some tests around the RandomShuffle extension method too, if possible, although I appreciate testing that algorithm probably won't be terribly easy.

adrianbanks:
It might be good to put the logic for the shuffle into another comparer (e.g. `RandomOrderSessionDisplayModelComparer`). You can then call sort with a different implementation of the comparer to get the difference in behaviour.

alastairs:
I was thinking that too, but held off suggesting it because it would require a quite different algorithm to achieve the random ordering: a Comparer only ever has two instances to look at, whilst @SimonPStevens' implementation hops about randomly over the entire list.

SimonPStevens:
@alastairs I'll add some tests for the extension, but I can't think of any good way to test the session controller change. Because it's random it's possible that the output would be ordered in exactly the same way, so there's nothing to assert in the test. Unless you just want some general tests around the session controller? - I'll do that.

adrianbanks:
@alastairs Fair point, although I do think it (with a different random sorting algorithm) would make the implementation a bit cleaner.

SimonPStevens:
@adrianbanks This algoithm is O(n), any algorithm based on a comparer is going to be less efficient, and personally I couldn't come up with one that didn't seem to have bias (I'm no statistician though). I did start out that way initially.

alastairs:
> @alastairs I'll add some tests for the extension, but I can't think of any good way to test the session controller change. Because it's random it's possible that the output would be ordered in exactly the same way, so there's nothing to assert in the test. Unless you just want some general tests around the session controller? - I'll do that.

Hmmm, that's why I thought the extension would be harder to test. Ideally some controller tests that say "if CanVote() use RandomShuffle" and "if !CanVote() use CustomAlphabeticalSort", just to test the appropriate methods were called in the right circumstances.

adrianbanks:
I'm ok with this going in - it just reads a bit odd:

```
if (X)
{
    do Y
}
else
{
    do something that doesn't resemble Y
}
```

As you have said, there are performance reasons and it would be hard(er) to do another way.

SimonPStevens:
@alastairs Not having using nunit before, is there a inbuilt way to do test if the methods were called, or do I need to add interfaces, and mocks?

alastairs:
Not in NUnit, no; you'll have to use mocks. We're using NSubstitute as our mocking library, documentation is [here](http://nsubstitute.github.io/).

adrianbanks:
@SimonPStevens For an example, take a look at line 62 of `ResetPasswordEmailSender_Should` in https://github.com/dddeastanglia/DDDEastAnglia/pull/129/files.

SimonPStevens:
@alastairs @adrianbanks Guys, I'm running out of time to do this (I'm going out soon). I've done some tests for the random sort algorithm, but I'm having problems with the session controller tests because the session controller is using the datacontext directly, so I'm going to have to switch that out too to use your messagebus. This might have to wait.

SimonPStevens:
@alastairs @adrianbanks I gave up on the tests for the session controller. They can wait until after the architecture meeting next week as there's too many changes needed to do that (and I haven't been involved enough with the code so far to know what direction to take on my own).

I've refactored the sorting so the decision on which algorithm to use is testable, but it all seems like overkill to essentially test what is one boolean if statement - perhaps I've missed something obvious here.

alastairs:
@SimonPStevens No worries, this is fine. I never meant for you to write tests for the whole `SessionController`, just the new behaviour you'd introduced. The rationale is that we have a lot of code not currently under test, and so I want to avoid adding new code that is untested as far as possible (we've already caught some flak from people in the community about the lack of tests and other good/best practices). 

What you've added here in the way of tests looks good. Apologies for the stressing you out!

PR #141, opened 2013-04-29
Make the voting buttons consistently styled like the other action items

adrianbanks:
:warning: **Do not merge yet - not ready**

The voting button had its own style, whereas it should have been styled the same as the other action items on a session.

PR #145, opened 2013-04-30
Fix #141: fix cookie end date

alastairs:
Set the cookie expiry date to be the date voting closes: 24 May.

I'm now not convinced this is the correct change to make, but hopefully another pair of eyes will be able to say one way or the other.


alastairs:
Verified on dev build that the voting button works...

Issue #127, opened 2013-04-25
About page needs updating with information on voting


alastairs:
Not sure what should be included exactly, but we should probably provide some information for newbies.

alastairs:
Ideally this would be done before voting opens, but in reality there are more pressing things to fix, and this could be delayed beyond the opening of voting by few days.

alastairs:
Seems a bit late to be adding this now.

PR #146, opened 2013-04-30
Add some descriptive text about voting

adrianbanks:


alastairs:
:+1: Looks good. Will do the manual merge now. There's also #127 to consider.

adrianbanks:
Yes, we need to do #127, but I thought this was a bit more obvious to people visiting the site.

PR #147, opened 2013-04-30
Fix wrapping of new message on Sessions page.

alastairs:
It was being forced to wrap by the large bottom margin on the Search box. I fixed it by dropping this value by 10px; we can revisit another time if needed.


adrianbanks:
I think that works. Can't remember why the margin was so big in the first place, but it doesn't seem necessary now.

PR #148, opened 2013-04-30
Fix the vote count on the admin page

adrianbanks:
Also adds a link to the menu (only visible to admin users).


alastairs:
Ok, I'm going to merge this, but these things need addressing post-merge:
- **TESTS!** (or rather the lack of them :unamused:)
- I think specifying `AdminHome` as the default controller should mean that most of the short-hand Action methods should work now (e.g., `Html.ActionLink()`) without specifying the controller. I.e. line 64 of the admin area's _Layout.cshtml.

PR #155, opened 2013-05-02
Fix #134: Introducing DI

### LINKED ISSUE #134: Introduce a DI container
It feels like we need one now that the application is getting bigger. I sure could have used one for the reset password change.

ninjaferret:
I've introduced Ninject into the project with some of the Ninject.Extensions.Conventions to automatically bind the right things. 

This means things like the Factory.cs can be removed (as well as some other unused files).

The ISortAlgorithm caused a bit of a problem because the two implementations needed to be bound to the two parameters on the SessionSorter so I bound each one specifically.


DDDEastAnglia/DataAccess/EntityFramework/Builders/Calendar/CalendarEntryBuilder.cs
```diff
@@ -6,19 +6,33 @@ namespace DDDEastAnglia.DataAccess.EntityFramework.Builders.Calendar
 {
     public class CalendarEntryBuilder : IBuild<CalendarItem, CalendarEntry>
```
alastairs, line 7: This class looks like it would be better implemented via the Abstract Factory pattern now, with Ninject "deciding" which implementation is created by injecting the appropriate concrete implementation to the requesting service.

DDDEastAnglia/DataAccess/EntityFramework/Builders/Calendar/CalendarEntryBuilder.cs
```diff
@@ -6,19 +6,33 @@ namespace DDDEastAnglia.DataAccess.EntityFramework.Builders.Calendar
...
+            return CreateSingleTimeEntry(item) ?? CreateTimeRangeEntry(item) ?? new NullCalendarEntry();
+        }
 
-        public CalendarEntryBuilder(IBuild<CalendarItem, SingleTimeEntry> singleTimeEntryBuilder,
-                                    IBuild<CalendarItem, TimeRangeEntry> timeRangeBuilder)
+        private CalendarEntry CreateSingleTimeEntry(CalendarItem item)
         {
-            _singleTimeEntryBuilder = singleTimeEntryBuilder;
-            _timeRangeBuilder = timeRangeBuilder;
+            if (item.EndDate.HasValue)
```
alastairs, line 16: I'm not 100% sure this is a valid concern of a factory class. At the very least it's a Law of Demeter violation. Perhaps there should be different types of `CalendarEntry` (open-ended vs. time-boxed) that can be iterated over by a Visitor or something to be processed appropriately.

DDDEastAnglia/DataAccess/MessageBus/SimpleMessageBus.cs
```diff
@@ -8,13 +8,12 @@ public class SimpleMessageBus : IMessageBus
...
         private readonly Dictionary<Type, IHandle> _handlers = new Dictionary<Type, IHandle>();
 
-        public SimpleMessageBus()
-        {
-            
-        }
-
         public SimpleMessageBus(IEnumerable<IHandle> handlers)
         {
+            if (handlers == null)
```
alastairs, line 13: This is the constructor, right? You should throw an `ArgumentNullException` here if so: it looks like the collection of handlers is a required dependency of this class.

DDDEastAnglia/Helpers/MvcHelperModule.cs
```diff
@@ -0,0 +1,31 @@
...
+        public override void Load()
+        {
+            Kernel.Bind(from => from.FromThisAssembly()
+                                    .SelectAllClasses()
+                                    .InNamespaceOf<IControllerInformationProvider>()
+                                    .Excluding<AlphabeticalSort>()
+                                    .Excluding<RandomSort>()
+                                    .BindDefaultInterfaces()
+                                    .Configure(binding => binding.InSingletonScope()));
+            Kernel.Bind<ISortAlgorithm>()
```
alastairs, line 22: This is slightly clumsy. I think it can be improved through the use of an Abstract Factory, as described above for a similar situation.

DDDEastAnglia/DataAccess/MessageBus/SimpleMessageBus.cs
ninjaferret, line 13: It is possible to register handlers via the Register method, I guess I could put a default constructor in

alastairs, line 13: No no no no, there should only be one constructor. If you don't want to register handlers at construction, an empty enumerable should be passed in.

DDDEastAnglia/DataAccess/EntityFramework/Builders/Calendar/CalendarEntryBuilder.cs
ninjaferret, line 16: There are multiple different types of `CalendarEntry` (the `SingleTimeEntry` and the `TimeRangeEntry`), did you mean `CalendarItem`? That would mean a change to the database or at least the mappings on the database to produce different classes. 

I was starting to work on specific builders for each type and have the 'CalendarEntryBuilder' take in specific builders for each type and return the first non-null value.

alastairs, line 16: Ok, could we discuss it before you dive down that hole? I possibly haven't understood the design you've got in place here.

adrianbanks:
Can this be merged? It would be very useful to have this in the codebase.

alastairs:
Looks like there are still a couple of things to tidy up.

alastairs:
@ninjaferret Please could you rebase onto latest production and I'll merge this in.

PR #159, opened 2013-05-06
Voting stats

adrianbanks:
I've added a dashboard to the admin area that shows lots of voting stats.

This pull request isn't finished yet, but I'd like some feedback on it, particularly how I test some of it, as most of the code is either javascript to produce charts or touches the database to run queries.


alastairs:
Mostly this looks ok so far. I'm a bit limited in my feedback, because I don't have much sample data in my database at all (added one session and vote). This has, however, turned up a few issues with robustness: most of the calculations you're doing throw an exception when there are no votes or sessions, rather than displaying an empty graph, or a message saying there are no votes. A possible alternative is to do a check up front whether the necessary data is available and disable the option if not. (Votes per day/hour work fine, as it happens.)

There also needs to be a mechanism for returning to the list of charts, beyond the nav bar and the browser's back button.

What about plotting "actual" statistics, such as means, medians, standard deviations, etc.?

The wells are quite large (and unnecessarily so) spread across the entire screen the way they currently are. Perhaps consider arranging them in a grid layout, rather than a series of rows?

The magnifying glass icon against individual IP addresses needs a tool tip to say what it will do.

DDDEastAnglia/Areas/Admin/Content/admin.css
```diff
@@ -57,3 +61,23 @@ a.btn, .dropdown-menu a, a.brand { text-decoration: none; }
     background-repeat: repeat-x;
     filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#ff013c01', endColorstr='#ff012301', GradientType=0);
 }
+
+.ui-progressbar {
```
alastairs, line 69: There are progress bars in Bootstrap that you can use: http://twitter.github.io/bootstrap/components.html#progress

DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    [Authorize(Roles = "Administrator")]
+    public class VotingController : Controller
+    {
+        private readonly DataProvider dataProvider;
+
+        public VotingController()
+        {
+            var connectionStringSettings = ConfigurationManager.ConnectionStrings["DDDEastAnglia"];
+            dataProvider = new DataProvider(connectionStringSettings.ConnectionString);
```
alastairs: What is a `DataProvider`?

DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    TotalVotes = dataProvider.GetTotalVoteCount(),
+                    NumberOfUsersWhoHaveVoted = dataProvider.GetnumberOfUsersWhoHaveVoted(),
+                    NumberOfDaysOfVoting = dataProvider.GetNumberOfDaysOfVoting(),
+                    NumberOfDaysOfVotingPassed = dataProvider.GetNumberOfDaysSinceVotingOpened(),
+                    NumberOfDaysOfVotingRemaining = dataProvider.GetNumberOfDaysUntilVotingCloses(),
+                    VotingStartDate = dataProvider.GetVotingStartDate(),
+                    VotingEndDate = dataProvider.GetVotingEndDate(),
+                };
+            return View(model);
```
alastairs, line 61: Re: testing. You can test that the correct view was rendered, and that it was rendered with the expected data. MvcContrib (or an equivalent library) will make this a bit easier. The same goes for the other actions, which do very similar things.

DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        [HttpPost]
+        [AllowCrossSiteJson]
+        public ContentResult LookupIPAddress(string ipAddress)
+        {
+            string hostName;
+
+            try
+            {
+                var ipHostEntry = Dns.GetHostEntry(ipAddress);
```
alastairs: If you put this behind an abstraction, it will be more testable. In fact, you could put the whole lot into that service, and just return a `ContentResult` with the host name returned from the service method.

DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+            catch
+            {
+                hostName = "[unknown]";
+            }
+
+            return Content(hostName);
+        }
+
+        public ActionResult VotesPerDay()
```
alastairs, line 101: Perhaps some clever routing would be more appropriate here, so you get `/admin/voting/day`, `/admin/voting/hour`, `/admin/voting/ipaddress`, etc.

DDDEastAnglia/Areas/Admin/Models/LeaderboardViewModel.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using DDDEastAnglia.VotingData.Models;
+
+namespace DDDEastAnglia.Areas.Admin.Models
+{
+    public class LeaderboardViewModel
+    {
+        public int HighestVoteCount{get;set;}
+        public IList<LeaderBoardSession> Sessions{get;set;}
```
alastairs: Should this be a `SessionLeaderBoard` (or perhaps a `SessionLeaderBoardEntry`)? `LeaderBoardSession` sounds a bit Yoda-speak.

DDDEastAnglia/Areas/Admin/Models/VotesPerDateTimeViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Linq;
+using DDDEastAnglia.VotingData.Models;
+
+namespace DDDEastAnglia.Areas.Admin.Models
+{
+    public class VotesPerDateTimeViewModel
+    {
+        public IList<DateTimeVoteModel> Votes{get;set;}
+
+        public string GetChartData()
```
alastairs: View Models are more usually POCOs, and so shouldn't have any behaviour. Instead, this should be done in the controller.

DDDEastAnglia/Areas/Admin/Views/Voting/Index.cshtml
```diff
@@ -0,0 +1,84 @@
...
+
+@section scripts
+{
+    <script>
+        $(function() {
+            $("#progressbar").progressbar({
+                max: @Model.NumberOfDaysOfVoting,
+                value: @Model.NumberOfDaysOfVotingPassed
+                });
+            $("#progresslabel").text("@((int) (Model.NumberOfDaysOfVotingPassed * 1.0f / Model.NumberOfDaysOfVoting * 100)) %");
```
alastairs: There shouldn't be logic in the view. This calculation should be done in the controller and passed to the view on the view model.

DDDEastAnglia/Helpers/JavasctiptDateTimes.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace DDDEastAnglia.Helpers
+{
+    public static class JavasctiptDateTimes
+    {
+        public static long GetJavascriptTimestamp(this DateTime input)
+        {
+            TimeSpan span = new TimeSpan(DateTime.Parse("1/1/1970").Ticks);
```
alastairs: `timeSinceEpoch` might be a better name for this variable?

DDDEastAnglia/VotingData/DataProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+using DDDEastAnglia.VotingData.Models;
+using DDDEastAnglia.VotingData.Queries;
+
+namespace DDDEastAnglia.VotingData
+{
+    public class DataProvider
+    {
+        private readonly QueryRunner queryRunner;
+
+        public DataProvider(string connectionString)
```
alastairs: The class depends on the `QueryRunner`; it doesn't care about the connection string at all.

DDDEastAnglia/VotingData/DataProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                    var durationUnitVotingCloses = votingDates.EndDate - DateTime.Today;
+                    return (int) durationUnitVotingCloses.Value.TotalDays;
+                });
+        }
+
+        private T GetVotingDates<T>(Func<CalendarItem, T> callback)
+        {
+            using (var context = new DDDEAContext())
+            {
+                CalendarItem votingDates = context.Conferences.Include("CalendarItems").First().CalendarItems.Single(c => c.EntryType == CalendarEntryType.Voting);
```
alastairs: A bit of wrapping on this line would be useful.

DDDEastAnglia/VotingData/DataProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                return callback(votingDates);
+            }
+        }
+
+        public int GetnumberOfUsersWhoHaveVoted()
+        {
+            using (var context = new DDDEAContext())
+            {
+                return context.Votes.AsEnumerable().GroupBy(v => v.CookieId).Count();
```
alastairs: Do you need the `AsEnumerable()`? I thought you could go straight to the `GroupBy()` because the things exposed on the context were already enumerables.

DDDEastAnglia/VotingData/DataProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Linq;
+using DDDEastAnglia.DataAccess.EntityFramework;
+using DDDEastAnglia.DataAccess.EntityFramework.Models;
+using DDDEastAnglia.Domain.Calendar;
+using DDDEastAnglia.VotingData.Models;
+using DDDEastAnglia.VotingData.Queries;
+
+namespace DDDEastAnglia.VotingData
+{
+    public class DataProvider
```
alastairs: This class doesn't feel like it hangs together all that well. Whilst objects without behaviour have their place (at the edges of an application), an object with lots of behaviour and no state can imply a low level of cohesion, and I think that is the case here. The methods exposed by this class don't present a cohesive interface; instead it seems to be more of a dumping ground for all the methods you need in another layer of the application.

DDDEastAnglia/VotingData/Queries/DistinctIPAddressQuery.cs
```diff
@@ -0,0 +1,40 @@
...
+                return @"
+SELECT IPAddress, COUNT(IPAddress) AS VoteCount FROM Votes GROUP BY IPAddress ORDER BY COUNT(IPAddress) DESC";
+            }
+        }
+
+        public IQueryResultObjectFactory<IPAddressModel> ObjectFactory
+        {
+            get
+            {
+                return new IPAddressModelFactory();
```
alastairs, line 24: Er, what?

DDDEastAnglia/VotingData/Queries/LeaderBoardQuery.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public IQueryResultObjectFactory<LeaderBoardSession> ObjectFactory
+        {
+            get
+            {
+                return new LeaderBoardSessionFactory();
+            }
+        }
+
+        private class LeaderBoardSessionFactory : IQueryResultObjectFactory<LeaderBoardSession>
```
alastairs: This looks like it's trying to be an Abstract Factory, but you seem to have taken a weird approach to implementing it...?

alastairs:
There is also a couple of small typos (incorrect casing, spelling mistakes) in some of your class names.

Re: JavaScript testing. A start would be to pull more of the JavaScript out of the pages and into separate files so they can be at least included into a test suite (e.g., QUnit). Better still would be to also use a framework that makes the production JavaScript code testable, such as [Knockout](http://knockoutjs.com) or [AngularJS](http://angularjs.org), or any one of the other MV\* JavaScript frameworks.

DDDEastAnglia/Areas/Admin/Models/VotesPerDateTimeViewModel.cs
adrianbanks: I did have that initially, but moved to this approach for sanity reasons.

To plot the charts requires data to be in the form of a javascript variable:

```
var data = [{
                color: "#C00",
                data: [[1367366400000, 1104], [1367452800000, 240], [1367539200000, 113], [1367625600000, 62], [1367712000000, 22], [1367798400000, 0], [1367884800000, 0], [1367971200000, 0], [1368057600000, 0], [1368144000000, 0], [1368230400000, 0], [1368316800000, 0], [1368403200000, 0], [1368489600000, 0], [1368576000000, 0], [1368662400000, 0], [1368748800000, 0], [1368835200000, 0], [1368921600000, 0], [1369008000000, 0], [1369094400000, 0], [1369180800000, 0], [1369267200000, 0], [1369353600000, 0], [1369440000000, 0]]
            }];
```

Trying to output that in the razor is nigh-on impossible as you need to get the correct javascript bracketing, which razor won't let you do easily when mixed with stuff from the model.

DDDEastAnglia/VotingData/DataProvider.cs
adrianbanks: > it seems to be more of a dumping ground for all the methods you need

At the moment it is. I needed some way of querying for data, and didn't want to wire in calls any particular way of doing it as there has been talk of removing EF. This class just abstract all of those calls, with the intention of refactoring it later into something that makes more sense.

DDDEastAnglia/Areas/Admin/Models/VotesPerDateTimeViewModel.cs
alastairs: Sounds like using JSON.NET (should already be included IIRC) to convert the C# object to the JSON representation might help you here, then: this method is, after all, manually serializing an object to JSON. It might also help you re-think the object structure to fix the Law of Demeter violations you have in the `foreach` loop (e.g., `item.Vote.Date`).

DDDEastAnglia/VotingData/DataProvider.cs
alastairs: > [I] didn't want to wire in calls any particular way of doing it as there has been talk of removing EF.

To be honest, I wouldn't let that stop you. Given there's less than two months until the conference itself now, we're probably better off treating this as a v1 site and sticking with EF, and then re-writing for next year.

DDDEastAnglia/Areas/Admin/Models/VotesPerDateTimeViewModel.cs
adrianbanks: Can you serialise using JSON.NET and have it omit the field names? That is what would be required to use it. I cannot find any docs that say it supports this.

alastairs: You would need a C# object in the same format as the JSON you're outputting: i.e., a collection of an object with two fields (color and data), where color is a string and data is a collection of collections. This seems appropriate to me, as it is the model the view requires. 

Which field names do you wish to omit?

adrianbanks: The data part is in the format `timestamp, numberOfVotes`. To hold this in a class, I would need something like:

```
public class Whatever
{
    public int Timestamp{get;set;}
    public int NumberOfVotes{get;set;
}
```

This gets output to JSON as:

```
{"Timestamp":1367366400000, "NumberOfVotes":1104}
```

To make it work, it would need to be serialised as:

```
[1367366400000, 1104]
```

This uses no field names, and square braces instead of curly ones.

alastairs: Both values are ints, so you could use a two-item array (or other collection) instead to achieve the output JSON you require. It's not as nice, but it's what the view requires...

DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        private readonly DataProvider dataProvider;
+        private readonly DnsLookup dnsLookup;
+
+        public VotingController()
+        {
+            var connectionStringSettings = ConfigurationManager.ConnectionStrings["DDDEastAnglia"];
+            var queryRunner = new QueryRunner(connectionStringSettings.ConnectionString);
+            dataProvider = new DataProvider(queryRunner);
+            dnsLookup = new DnsLookup();
```
alastairs: This doesn't buy you much unless you inject the `DnsLookup` service class into the `VotingController`: it's still as difficult to test (requiring a full DNS stack) as it was before.

That said, it does improve the cohesion of the `VotingController` a bit, and of course `DnsLookup` is fully SRP-compliant.

adrianbanks:
I think this is ready to merge now. I still have ideas for more stats, but they can be added later.

PR #163, opened 2013-05-11
Update font awesome to v3.1.1

adrianbanks:


alastairs:
FontAwesome was added via NuGet, but I don't see any changes to the packages.config file here. Was this update done manually?

adrianbanks:
Didn't realise it was added via NuGet. The new version isn't on NuGet (it looks like the person who made the NuGet package isn't the author/owner of Font Awesome so it might be a while before it is), but I could really do with two of the new icons from it.

If it was added using NuGet, why are some bits of it checked in?

alastairs:
Which are the icons you need?

> If it was added using NuGet, why are some bits of it checked in?

Because the items (e.g. font files) have to be included in the project file in order to get deployed to Azure and I don't like having broken references in the project.

adrianbanks:
I need the new expand/collapse icons for the voting stats. Didn't want to update FA in this branch though.

> Because the items (e.g. font files) have to be included in the project file in order to get deployed to Azure and I don't like having broken references in the project.

How are they broken once NuGet has restored the packages?

alastairs:
> I need the new expand/collapse icons for the voting stats.

Can you make do with the existing expand/collapse icons _for now_ until the NuGet package has been updated?

> How are they broken once NuGet has restored the packages?

They aren't, but NuGet doesn't automatically restore the packages. I also thought that, given that the files are either small or text or both, that it was ok to check them in.

adrianbanks:
That was the problem - there aren't any existing expand/collapse icons. I've re-purposed two icons that sort of work, but the new expand/collapse ones are much more obvious as to what action you can perform.

Based on the logic of why you've checked the files in, why don't we remove the NuGet package and manually update it? You can just drop the download over the top of what is checked in.

adrianbanks:
The NuGet package has now been updated, so I've updated via that.

PR #164, opened 2013-05-12
Add Eventbrite integration for registration

alastairs:
I have added two items from Eventbrite:
1. A countdown widget in the sidebar, to replace the existing event details. 
2. A registration page containing the registration form, linked from the nav bar.

Thanks to the caching we're already doing on the website, and the fact that the registration form is served from Eventbrite via an iframe, I think we will be able to do registration through our website rather than pointing people at Eventbrite. We can always use the Eventbrite URL as a fall-back.

### DIFF
--- a/DDDEastAnglia/Views/Shared/_Layout.cshtml
+++ b/DDDEastAnglia/Views/Shared/_Layout.cshtml
@@ -53,6 +53,7 @@
                             <li class="dropdown @Html.MenuState("Session")">@Html.ActionLink("Sessions", "Index", "Session")</li>
                             <li class="dropdown @Html.MenuState("Speaker", "Index")">@Html.ActionLink("Speakers", "Index", "Speaker")</li>
                             @*<li class="dropdown">@Html.ActionLink("Agenda", MVC.Agenda.Index())</li>*@
+                            <li class="dropdown @Html.MenuState("Home", "Register")">@Html.ActionLink("Register", "Register", "Home")</li>
                             <li class="dropdown @Html.MenuState("Home", "About")">@Html.ActionLink("New to DDD?", "About", "Home")</li>
                             <li class="dropdown @Html.MenuState("Home", "Venue")">@Html.ActionLink("Venue", "Venue", "Home")</li>
                             <li class="dropdown @Html.MenuState("Home", "Sponsors")">@Html.ActionLink("Sponsors", "Sponsors", "Home")</li>
--- a/DDDEastAnglia/Controllers/HomeController.cs
+++ b/DDDEastAnglia/Controllers/HomeController.cs
@@ -33,5 +33,10 @@ namespace DDDEastAnglia.Controllers
         {
             return View();
         }
+
+        public ActionResult Register()
+        {
+            return View();
+        }
     }
 }
--- a/DDDEastAnglia/DDDEastAnglia.csproj
+++ b/DDDEastAnglia/DDDEastAnglia.csproj
@@ -524,6 +524,7 @@
       <SubType>Code</SubType>
     </Content>
     <Content Include="Views\Shared\_SearchBox.cshtml" />
+    <Content Include="Views\Home\Register.cshtml" />
   </ItemGroup>
   <ItemGroup>
     <Folder Include="App_Data\" />
--- a/DDDEastAnglia/Views/Home/Register.cshtml
+++ b/DDDEastAnglia/Views/Home/Register.cshtml
@@ -1,0 +1,11 @@
+﻿@{
+    ViewBag.Title = "Register";
+}
+
+<h2>Register</h2>
+
+<p>
+    Registration will open at 1.30pm on Monday 3 June.
+</p>
+
+<div style="width:100%; text-align:left;"><iframe src="http://www.eventbrite.co.uk/tickets-external?eid=5821705879&ref=etckt&v=2" frameborder="0" height="214" width="100%" vspace="0" hspace="0" marginheight="5" marginwidth="5" scrolling="auto" allowtransparency="true"></iframe><div style="font-family:Helvetica, Arial; font-size:10px; padding:5px 0 5px; margin:2px; width:100%; text-align:left;"><a style="color:#ddd; text-decoration:none;" target="_blank" href="http://www.eventbrite.co.uk/r/etckt">Online Ticketing</a><span style="color:#ddd;"> for </span><a style="color:#ddd; text-decoration:none;" target="_blank" href="http://dddea2013.eventbrite.co.uk?ref=etckt">DDD East Anglia</a> <span style="color:#ddd;">powered by</span> <a style="color:#ddd; text-decoration:none;" target="_blank" href="http://www.eventbrite.co.uk?ref=etckt">Eventbrite</a></div></div>


DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -53,6 +53,7 @@
                             <li class="dropdown @Html.MenuState("Session")">@Html.ActionLink("Sessions", "Index", "Session")</li>
                             <li class="dropdown @Html.MenuState("Speaker", "Index")">@Html.ActionLink("Speakers", "Index", "Speaker")</li>
                             @*<li class="dropdown">@Html.ActionLink("Agenda", MVC.Agenda.Index())</li>*@
+                            <li class="dropdown @Html.MenuState("Home", "Register")">@Html.ActionLink("Register", "Register", "Home")</li>
```
adrianbanks, line 56: We shouldn't be showing the register link until you can actually register, just like we did for session submission and voting.

adrianbanks:
The countdown widget in the sidebar is quite ugly. It also makes a bit of a hash of displaying the venue details.

Do we really need the countdown?

DDDEastAnglia/Views/Shared/_Layout.cshtml
alastairs, line 56: I was thinking that too, but a) there's no way to register until the given date and time, and b) I can't work out how we do this. Is there anything in the site to do this already?

Don't we also need to turn off some of the other pages?

alastairs:
Hmm, yeah, I wasn't completely happy with how it turned out when I customised it with our colours, but the options are rather limited. 

I guess strictly speaking we don't _need_ the countdown, no, but I thought it was a nice feature to include.

DDDEastAnglia/Views/Shared/_Layout.cshtml
adrianbanks, line 56: Yes, the session submission and voting links appear/disappear based on dates. I think the registration should do the same.

adrianbanks:
> Hmm, yeah, I wasn't completely happy with how it turned out when I customised it with our colours, but the options are rather limited.

The formatting of the address doesn't look right. There's also some weird shading going on with the countdown number that makes it look weird.

> I guess strictly speaking we don't need the countdown, no, but I thought it was a nice feature to include.

It would be more useful to have a countdown to the start of registration that the start of the event.

One thing I've just noticed is that by replacing the text with the widget, we've lost the only mention of the end time of the event.

alastairs:
> The formatting of the address doesn't look right. There's also some weird shading going on with the countdown number that makes it look weird.

Agreed. See previous comment about limited customisation options. 

> The formatting of the address doesn't look right. There's also some weird shading going on with the countdown number that makes it look weird.

I disagree. It would be weird to countdown to registration opening and not the event: it would imply registration is a bigger event than the event itself. 

> we've lost the only mention of the end time of the event.

Fair point, I'd missed that.

adrianbanks:
For most attendees of the event, they want to know when registration opens so that they can sign up. Once they are signed up, some of them will probably not even come back to the site before the event, some will come back to get a pertinent piece of information (e.g. address, agenda). Having a countdown doesn't really add any value for them.

alastairs:
I've removed the countdown widget from the side bar.

DDDEastAnglia/Views/Shared/_Layout.cshtml
alastairs, line 56: The links in the nav bar are currently completely static, but I'll see what I can do.

alastairs, line 56: Ok, it's a no-go. In order to do this, I either need to add an `IsOpenForRegistration` property to every model _and_ supply a model to every single view, or pull out the nav bar into a separate controller with its own model, or put logic into the view. I'm not willing to do any of those.

adrianbanks, line 56: You can use a partial view for this. We do it for the sponsors in the side bar.

alastairs, line 56: No, I can't. Not without passing a model to the partial view (and thus having to pollute every other view), or putting lots of logic in the partial view. The problem is that it's rendered in the Layout page, and it requires some data to decide whether or not to display the link. The sponsors in the side bar are just static content.

adrianbanks, line 56: You can do it using a partial view and a call to `@{Html.RenderAction("Action", "Controller")}` from the shared Layout view. If your action returns a `PartialView()` instead of a `View()`, you can pass a model created in the controller's action method to the partial.

alastairs, line 56: Wouldn't that mean I would have to update _every single action_ currently defined, and any future actions added would also have to include this?

adrianbanks, line 56: No. The shared layout calls the action on the controller, which creates a model and passes it to a partial view. The partial view takes the model and creates its output which is returned to the shared view for rendering. Nothing else needs to change.

We should possibly do the whole menu this way, so that we can show/hide individual links throughout the timeline of the event. For instance, the agenda link needs to appear at some point, and the admin link should only show to admin users without the check being done in the view itself like it currently is.

alastairs, line 56: Which controller??

adrianbanks, line 56: > The shared layout calls the action on the controller,

If you are referring to this controller, you specify it in the `Html.RenderAction` call.

alastairs, line 56: **Which** controller?

My understanding is as follows: the routing engine receives a request and resolves the URL to an action on a controller. The controller then does its processing and returns an appropriate `ActionResult`. If that `ActionResult` is a `ViewResult`, the Layout must include a call to `RenderBody()` to "insert" the view in the `ViewResult` at the appropriate point in the page. It doesn't make sense for an action to be called by the Layout, because it might not return a ViewResult: what happens if your action returns a `JsonResult`, for example? It makes no sense to render that on a page.

adrianbanks, line 56: Have a look at [this](https://github.com/adrianbanks/dddeastanglia/commit/38e202fa9b6f68f9909b755fc3f0692d45ed3c47) commit that is a working example.

alastairs, line 56: Right, that makes more sense now, thanks. Will get that sorted tomorrow.

alastairs, line 56: I've been looking at this for the last half-hour or so. It's a hell of lot of boilerplate for one line of HTML, plus the necessary associated tests, which are hard to write because there are abstractions missing. In order to get this change in, I will need to:
- Introduce MVCContrib.TestHelper to the tests project (fine)
- Add tests for the new controller (fine)
- Add a DateTimeProvider abstraction (we don't have one of these already?)
- Add an abstraction over at least one, probably two, of the Entity Framework entities (Conferences and CalendarItems) so they can be injected into the controller.
- More stuff that I haven't got to yet

Given that the page currently says "Registration will open at 1.30pm on Monday 3 June", I really do think displaying the registration page as-is is a sufficient and pragmatic implementation of the feature. I don't see that there's a problem displaying this page when it offers useful information (date and time that registration opens), as well as a URL that visitors can bookmark for the day.

adrianbanks, line 56: I think it is worth doing, as if we are making a site that can handle future events, we should be rendering the some of the other menu items dependent on availability due to date. If we are not making a site that can handle future events, then we are over-complicating the current site.

What you have said needs doesn't sound like it will be that much code, and doesn't look like as much effort as has already gone into other work done on the site. If you are adamant that this is too hard, let's merge this to get the integration of EventBrite, then address this particular issue separately.

As an aside, you don't need MVCContrib.TestHelper to unit test the views. I've tested controllers and views on the voting branch without it.

alastairs, line 56: I expect to be able to reuse bits of/most of the site for future events, but I don't necessarily expect us to use Eventbrite for registration in future. As such, I view this as a throw-away implementation of the feature that is good enough for v1 but will be replaced by something better for v2 when we have more time. 

> As an aside, you don't need MVCContrib.TestHelper to unit test the views. I've tested controllers and views on the voting branch without it.

Agreed you don't _need_ it, but it does make the tests nicer.

adrianbanks, line 56: > but I don't necessarily expect us to use Eventbrite for registration in future.

We are talking about showing the registration link in the nav bar here, not the implementation of the page itself. That will surely be needed in the future, for both the registration link and the other temporal pages?

alastairs, line 56: Yes, it will be needed in future, but I don't think it's needed now. See previous comment about the current implementation showing useful information before registration opens.

adrianbanks:
It isn't much effort to implement this to pull the information that is already in the database for the registration. See [this](https://github.com/adrianbanks/dddeastanglia/commit/4f3c302858de0ae59c9335308e6d21c81e511866) commit (it obviously needs some tests).

alastairs:
It is a disproportionate amount of work in comparison with the work required for this feature (18 LOC), and smells like feature-creep to me. I will merge this, and the nav bar work can be tackled as a separate item.

adrianbanks:
I still don't see what the problem is, especially when we've had this situation before and done something similar, but ok - merge this and we'll tackle the nav bar in a separate pull request (I might as well do it as I've got 90% of the solution on a branch in my repo).

PR #168, opened 2013-05-15
Extract the navigation bar into a separate (partial) view

adrianbanks:


DDDEastAnglia.Tests/Controllers/NavigationBarControllerShould.cs
```diff
@@ -0,0 +1,78 @@
...
+        public void ThrowAnException_IfConstructedWithANullConferenceRepository()
+        {
+            Assert.Throws<ArgumentNullException>(() => new NavigationBarController(null));
+        }
+
+        [Test]
+        public void IndicateThatTheUserIsAnAdministrator_WhenTheUserIsInTheAdministratorRole()
+        {
+            var conference = Substitute.For<IConference>();
+            conference.CanPublishAgenda().Returns(true);
```
alastairs: Why does this test care about whether the agenda can be published or not?

adrianbanks: Dunno, I'll remove it.

DDDEastAnglia.Tests/Controllers/NavigationBarControllerShould.cs
```diff
@@ -0,0 +1,78 @@
...
+            return CreateController(conference, user);
+        }
+
+        private NavigationBarController CreateController(IConference conference, IPrincipal user)
+        {
+            var conferenceRepository = Substitute.For<IConferenceRepository>();
+            conferenceRepository.GetByEventShortName(Arg.Any<string>()).Returns(conference);
+
+            var controllerContext = Substitute.For<ControllerContext>();
+            controllerContext.HttpContext.User.Returns(user);
```
alastairs, line 73: IIRC, I think this stuff is also provided by `MvcContrib.TestHelper`. There's definitely a helper ControllerFactory for setting the HttpContext, etc., in a nicer way. 

Happy to leave it as-is for now, though.

DDDEastAnglia.Tests/Controllers/NavigationBarControllerShould.cs
```diff
@@ -0,0 +1,78 @@
...
+            var controller = CreateController(conference, user);
+
+            var result = (PartialViewResult) controller.Index();
+
+            var model = (NavigationMenuViewModel) result.Model;
+            Assert.IsTrue(model.IsUserAnAdministrator);
+        }
+
+        [Test]
+        public void IndicateThatRegistrationIsOpen_WhenTheEvenReportsThatRegistrationIsOpen()
```
alastairs: Typo: Even => Event. (Also in the next test.)

Should these be something like "DisplayTheRegistrationLink", rather than "IndicateThatRegistrationIsOpen"? If you were writing a component that sent some notification, the "Indicate" wording would make sense, but given that the SUT is a controller that decides whether or not to render links to the user it doesn't read as explicitly as it might.

DDDEastAnglia.Tests/Controllers/NavigationBarControllerShould.cs
```diff
@@ -0,0 +1,77 @@
...
+            var controller = CreateController(conference, user);
+
+            var result = (PartialViewResult) controller.Index();
+
+            var model = (NavigationMenuViewModel) result.Model;
+            Assert.IsTrue(model.IsUserAnAdministrator);
+        }
+
+        [Test]
+        public void IndicateThatRegistrationIsOpen_WhenTheEventReportsThatRegistrationIsOpen()
```
alastairs, line 36: Should these be something like "DisplayTheRegistrationLink", rather than "IndicateThatRegistrationIsOpen"? If you were writing a component that sent some notification, the "Indicate" wording would make sense, but given that the SUT is a controller that decides whether or not to render links to the user it doesn't read as explicitly as it might.

DDDEastAnglia/Controllers/NavigationBarController.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public ActionResult Index()
+        {
+            bool userIsAdmin = User.IsInRole("Administrator");
+            var conference = conferenceRepository.GetByEventShortName(DefaultEventName);
+            bool isRegistrationOpen = conference.CanRegister();
+            bool isAgendaPublished = conference.CanPublishAgenda();
+
+            var model = new NavigationMenuViewModel
```
alastairs, line 30: You could inline the setting of the view model properties, given that the variables being assigned to them are named almost identically.

DDDEastAnglia.Tests/Controllers/NavigationBarControllerShould.cs
adrianbanks, line 36: I did think that as I was writing the test, but I'm not actually testing that a link is displayed - I'm testing that a property is set of the view model.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
```diff
@@ -0,0 +1,30 @@
+﻿@model DDDEastAnglia.Models.NavigationMenuViewModel
+
+<div class="nav-collapse">
+    <ul class="nav">
+        <li class="dropdown @Html.MenuState("Home", "Index")">@Html.ActionLink("Home", "Index", "Home")</li>
```
alastairs, line 5: Now that we have a navbar controller, it seems like this menu state logic may be better placed on the controller than in an extension method? Just a thought.

DDDEastAnglia.Tests/Controllers/NavigationBarControllerShould.cs
alastairs, line 36: Fair point. Ok :+1:

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 5: Didn't realise that was an extension method :smile: 

I'll take a look tomorrow.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
```diff
@@ -0,0 +1,30 @@
...
+
+        <li class="dropdown @Html.MenuState("Home", "About")">@Html.ActionLink("New to DDD?", "About", "Home")</li>
+        <li class="dropdown @Html.MenuState("Home", "Venue")">@Html.ActionLink("Venue", "Venue", "Home")</li>
+        <li class="dropdown @Html.MenuState("Home", "Sponsors")">@Html.ActionLink("Sponsors", "Sponsors", "Home")</li>
+        <li class="dropdown @Html.MenuState("Home", "Team")">@Html.ActionLink("Team", "Team", "Home")</li>
+        <li class="dropdown @Html.MenuState("Home", "Contact")">@Html.ActionLink("Contact", "Contact", "Home")</li>
+                            
+        @if (Model.IsUserAnAdministrator)
+        {
+            <li class="dropdown">@Html.ActionLink("Admin", "Index", new { area = "Admin", controller = "AdminHome" })</li>
```
alastairs, line 27: Might be nice for this button to display on the right of the nav bar, to keep it separate from the others. You can use `.pull-right`, and perhaps a separate `.btn-group` too, to achieve this. There might also be a nav bar separator that you can use instead, can't remember.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 5: I've taken a look at this and there doesn't seem to be a nicer way of doing it without having a property for each menu item on the model.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 27: Can't seem to make anything work here that looks nicer than what we have. The bootstrap styles don't play nice with each other - the link either vanishes completely (it has a height of 0px) or doesn't pull right.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
alastairs, line 5: What about having a separate model type that represents a nav bar link and exposing a collection of those on the `NavigationMenuViewModel`?

DDDEastAnglia/Views/NavigationBar/Index.cshtml
alastairs, line 27: Hmmm, that doesn't sound right: the login drop-down does this exact thing. Did you try defining a separate `.btn-group`?

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 5: Not a bad idea. How would I get around hard-coding the text for the links in code instead of the view?

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 27: Yes. The problem is that to float it right, the admin menu item has to live outside of the menu (i.e. in the main `_Layout` view), otherwise it floats to the right of the menu and is in the same place as it currently is.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
alastairs, line 5: I don't think you could. What bits of information does a nav bar link need to render?
- Text
- URL
- Selected state
- Visibility

These all jump out immediately from the code below, but there might be other bits I've missed.

Is it a problem to hard-code the link text in code rather than the view?

adrianbanks, line 5: Yes, those four sound correct.

> Is it a problem to hard-code the link text in code rather than the view?

Not a problem as such, but loses the flexibility of being able to tweak them without recompilation.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
alastairs, line 27: Could you not just add an extra `<div>` into this view for the admin menu item? You could perhaps also pull the login drop-down `<div>` into this view too, and the nav-collapse button, so it's the view for all the nav bar rather than just the navigation buttons.

Dunno, I'm not that fussed overall, I'm just conscious that the nav bar is starting to get quite cluttered and that the admin button a) doesn't really belong with the other buttons there, and b) can be not-completely-obvious to find. (By "not-completely-obvious", I just mean "requires a moment's thought", not "tricky".)

DDDEastAnglia/Views/NavigationBar/Index.cshtml
alastairs, line 5: True. It seems like a small price to pay, though. You could perhaps do something funky with an extension method on `string` that takes the model or the current link...

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 27: Can't seem to get it to work well at all. This is the best I've had so far:

capture

which still isn't correct as the link on the button is underlined for some reason. It also breaks the line wrapping when the window is resized:

capture1

alastairs, line 27: Ugh. Ok, let's leave it and revisit it another time. Perhaps file an issue for it.

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 5: I've just had a go at this and it is **way** too much effort to be worth it. I am currently up to 6 new classes/interfaces just to wire it in in code. I've yet to wrap decent testing around it, but I think I need to add more wrappers around the framework to test it.

The problem stems from the fact that the link generation and the currently selected state rely heavily on the framework (`ControllerContext`). I've also lost R#'s check that the actions exist now that they are called in code instead of in the view.

alastairs, line 5: I think you can call the UrlHelper on the controller, which would be a R#-friendly approach to building the URL. That's kind of a detail.

If it gives you any inspiration, I did something similar to render a dynamic list of links on another site. I don't have the site running live to demo, but take a peek at https://github.com/alastairs/cgowebsite/tree/develop/src/CGO.Web/Controllers (specifically, `SideBarController`, the `ConcertsSideBar`, and the `ConcertsSideBarFactory`). and the associated models in https://github.com/alastairs/cgowebsite/tree/develop/src/CGO.Web/Models.

If it's still too much code, I'll do the merge tomorrow and we can revisit it another time (or forget it :smile:).

adrianbanks, line 5: My case is slightly different I think in that I have to take the _current_ `ControllerContext` for the action being called to determine the active menu item. This muddies the water. I think I have something close to working, so I'll push it to another branch and let you take a look.

alastairs, line 5: I'm querying the current `ControllerContext` in my other site to determine which SideBar to render (if any): https://github.com/alastairs/cgowebsite/blob/develop/src/CGO.Web/NinjectModules/SideBarModule.cs  Pretty certain you could adapt that code to your purposes?

Do you have a link for your other branch?

DDDEastAnglia/Views/NavigationBar/Index.cshtml
alastairs, line 27: Hmm, a thought: I wonder if the wrapping is broken anyway now that we have more links on the nav bar?

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 5: > Do you have a link for your other branch?

https://github.com/adrianbanks/DDDEastAnglia/tree/extracting-menu-v2

DDDEastAnglia/Views/NavigationBar/Index.cshtml
adrianbanks, line 27: Possibly. I'll check tonight.

adrianbanks, line 27: The wrapping is currently fine when the admin link is not there. It is broken when it is showing. The same thing will happen when the Register link appears.

alastairs:
Superseded by #176.

PR #169, opened 2013-05-15
Allow the deployment script to be run more than once

adrianbanks:
This also fixes an issue with inserting specific identities into tables.


alastairs:
Ta, I'd noticed that issue as well.

PR #170, opened 2013-05-15
Fix the paths to font awesome

adrianbanks:
The new font awesome has moved the font path in relation to the rest of it, so the icons are broken.


alastairs:
The NuGet package moved the files around?

adrianbanks:
I can't quite figure out what's going on, as it appears that the `font` folder needs to be in the root of the site - not where the NuGet package put it. See if you have any more luck.

PR #173, opened 2013-05-15
Use the standard bootstrap table styles on the accommodation page

adrianbanks:


DDDEastAnglia/Content/Site.css
```diff
@@ -296,6 +296,10 @@ fieldset {
 .hotel-details.primary-hotel {
     border-color: #008600;
 }
+ 
+.hotel-details table {
+    width: auto;
```
alastairs, line 301: Do you even need this? The `.table` class sets the width to be 100%, which should be sufficient?

adrianbanks, line 301: Yes. Without it, the image that is floated right causes the table to wrap underneath the image leaving a big white hole within the section.

PR #176, opened 2013-05-22
Extract the navigation bar into a separate (partial) view (v2)

adrianbanks:
This supersedes #168 really as it takes that branch and treats the menu as a series of links.


DDDEastAnglia.Tests/NavigationMenu/MenuStateFactoryShould.cs
```diff
@@ -0,0 +1,26 @@
...
+            var factory = new MenuStateFactory();
+            Assert.Throws<ArgumentNullException>(() => factory.Create(null));
+        }
+
+        [Test]
+        public void ReturnAMenuState_WhenCalledWithANonNullRouteData()
+        {
+            var factory = new MenuStateFactory();
+            var menuState = factory.Create(new RouteData());
+            Assert.IsNotNull(menuState);
```
alastairs: An instance-of assertion would be more appropriate here, I think.

adrianbanks: Do you mean `Assert.That(menuState, Is.Not.Null)`? If so, I don't think it really matters.

DDDEastAnglia/Controllers/NavigationBarController.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            var conference = conferenceRepository.GetByEventShortName(DefaultEventName);
+            var links = new List<NavigationMenuLinkViewModel>
+                {
+                    CreateLink("Home", "Home", "Index"), 
+                    CreateLink("Sessions", "Session", "Index", () => !conference.CanPublishAgenda() && (conference.CanSubmit() || conference.CanVote())),
+                    CreateLink("Speakers", "Speaker", "Index"),
+                    CreateLink("Agenda", "Agenda", "Home", conference.CanPublishAgenda),
+                    CreateLink("Register", "Register", "Home", conference.CanRegister),
+                    CreateLink("About", "About", "Home"),
```
alastairs: This currently says "New to DDD?". Do we want to change it to "About"?

adrianbanks: I didn't change it intentionally.

DDDEastAnglia/Controllers/NavigationBarController.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                throw new ArgumentNullException("urlHelperFactory");
+            }
+            
+            this.conferenceRepository = conferenceRepository;
+            this.menuStateFactory = menuStateFactory;
+            this.urlHelperFactory = urlHelperFactory;
+        }
+
+        public ActionResult Index()
```
alastairs: This should be called `RenderMenu` or something similar, and should have the `ChildActionOnly` attribute applied.

DDDEastAnglia/Controllers/NavigationBarController.cs
```diff
@@ -0,0 +1,78 @@
...
+                    CreateLink("Team", "Team", "Home"),
+                    CreateLink("Contact", "Contact", "Home"),
+                    CreateLink("Admin", "AdminHome", "Index", () => User.IsInRole("Administrator"), new { area = "Admin" }),
+                };
+
+            var model = new NavigationMenuViewModel { Links = links };
+            return PartialView(model);
+        }
+
+        private NavigationMenuLinkViewModel CreateLink(string linkText, string controllerName, string actionName,
```
alastairs, line 66: For SRP, this should be in a separate class, but happy to leave as-is.

DDDEastAnglia.Tests/NavigationMenu/MenuStateFactoryShould.cs
alastairs: No, I mean check that the type is what you expect. `Assert.InstanceOf<MenuState>()`. You have mixed your assertion styles, but I was letting that slide as the individual test fixtures appear to be self-consistent.

adrianbanks: Ah, I see.

alastairs:
Please could you rebase onto latest `production`, then we'll get this merged.

adrianbanks:
Rebased onto production. There is still a conflict in `_Layout.cshtml` because you added the registration link in your pull request that has already been merged into staging. You can just ignore that change (in the conflicted file - take your other changes), as I have included the registration link in the change as well.

PR #177, opened 2013-05-22
Add legends to the voting stats charts

adrianbanks:
Also fixes a small bug whereby a user with a blank name shows with no name in the stats page.


alastairs:
Looks to add axis labels rather than a legend.

PR #178, opened 2013-05-23
Add Gibraltar logo

alastairs:
The image file came from Gibraltar as-is, and was tweaked by @adrianbanks to remove the background colour.


adrianbanks:
Can you replace the logo with this one where I've fixed the transparency?

gibraltar

alastairs:
Rebased onto latest production and updated the logo.

PR #179, opened 2013-05-23
Fix the transparency on the sponsor images

adrianbanks:
This involved turning two of them from jpg files to png files.


alastairs:
Is this necessary? The images display on a white background, after all, and it's not like we're going to change the background colour of the site for this year's conference.

adrianbanks:
It was to make the logos consistent, especially as we may need them for graphics elsewhere (fliers, slides, etc...). We'll obviously need higher resolution ones for banners and the like.

alastairs:
Hmmm, ok. I think slides are probably the only other place these could be reused: the standard resolution for printed materials is 300dpi, whereas these are all (correctly) 72dpi or so.

adrianbanks:
Well it's done now, and it does no harm to have them ready for the slides.

alastairs:
The Red Gate and JetBrains logos appear unchanged?

adrianbanks:
In the pull request yes, but put them on a background colour and I've removed some white artefacts around the edge of the anti-aliasing.

PR #182, opened 2013-05-27
Agenda

philpursglove:
PR to bring in the agenda page


philpursglove:
I wonder whether we should have a print stylesheet that hides the nav bar, sidebar and footer - any thoughts?

alastairs:
Sounds like a plan.

adrianbanks:
I think you need to pass a `null` as an additional last parameter into each `@Html.ActionLink()` call otherwise the links will be broken.

PR #184, opened 2013-05-29
Fix #135: Add the current commit SHA1 to the admin page

### LINKED ISSUE #135: Display the current commitish somewhere in the site.
I would suggest in the admin area. This would enable us to see where the live site is in the log.

adrianbanks:
This is only generated in the build system, so on a local machine it should be `[unknown]`.


alastairs:
This appears to be failing in the SSDT project: cannot find `Microsoft.SqlServer.TransactSql`. No idea right now why... [Build log](http://build.dddeastanglia.com/viewLog.html?buildId=718&buildTypeId=bt8&tab=buildLog)

alastairs:
Got it: it's the change to the solution file. If you look at the diff of the solution file on this branch, the database project is now being parsed correctly as a member of the solution. This isn't the case on staging: the database project has never been built on TeamCity as part of the staging branch (but was when it was on my feature branch) :worried: It's therefore one of the changes made to the database project _since the project was added_ that has added this extra binary dependency in the build.

I'll fix this up on a separate branch.

adrianbanks:
I wondered about the solution change, but didn't see how changing it to be correct could have caused the problem. Hadn't thought that it wasn't even building before.

alastairs:
It will be `[unknown]` on both the staging and production sites as well, because they are deployed separately from the build process. The other part of this change is to create a deployment package from the build that can be pushed up to staging and production.

PR #210, opened 2014-03-22
Add JsonConvertWrapper

philpursglove:
Adds JsonConvertWrapper to fix JsonConvert visibility weirdness on Phils machine.


alastairs:
Do we actually want this in the site...?

adrianbanks:
Not really, but it is hampering Phil from working on the site.

alastairs:
Yeah, so could we leave it as a temporary thing for @philpursglove today, rather than making it a permanent thing that is otherwise unnecessary?

philpursglove:
That's fine, but it implies the existence of another solution which is non-temporary.

alastairs:
Fixed without code changes.

PR #211, opened 2014-03-22
2014 updates

alastairs:
This updates the site for 2014. It removes all references to 2013, the Hauser Forum, etc., and "resets" the site to the pre-conference state: no sessions, speakers, agenda, etc. 

**Note:** This needs checking before merging - I haven't been able to test it at all as I don't currently have a database server against which I can run my debug site.


adrianbanks:
When running, this suffers from the "automatic migrations are disabled" problem you mentioned before.

alastairs:
Interesting. Looks like something funky's going on with the database then? We'll get your Simple.Data branch merged next...

adrianbanks:
The staging site still has 2013 remnant on it (speaker list, accommodation tab, the venue tab has the wrong details). Is this because of the staging database, or does it need to be altered before this goes to production?

alastairs:
Yes, the database needs poking to fix those things. The same will need to be done on production too.

PR #215, opened 2014-03-22
Upgrade the solution to .NET 4.5.1.

alastairs:
Upgrade to the latest version of .NET for all the new platform features.

See http://go.microsoft.com/fwlink/?LinkId=235367 for details of the changes made to the web.config file.

### DIFF
--- a/DDDEastAnglia.Tests/DDDEastAnglia.Tests.csproj
+++ b/DDDEastAnglia.Tests/DDDEastAnglia.Tests.csproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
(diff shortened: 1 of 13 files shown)


alastairs:
This is ready to merge. @adrianbanks - you might have some conflicts here and there with your replace-EF-with-SimpeData branch, but it shouldn't be anything major...

adrianbanks:
This is going to cause me problems. I only have VS2012 on my laptop, and even though I've installed .Net 4.5.1, VS insists that I don't have it installed and refuses to open the project.

alastairs:
Bum. That might be because I've been working in VS2013 all day and so the changes I've made support? 

(Do you have the option to upgrade to VS2013 on your laptop?)

adrianbanks:
> Do you have the option to upgrade to VS2013 on your laptop?

I will do, but I think Phil is also on VS2012. You could just upgrade the projects to VS2013, but not bump the framework version.

alastairs:
> You could just upgrade the projects to VS2013, but not bump the framework version.

There are no project file changes necessary to load the solution into VS2013. The point of the pull request is to bump the framework version so we can use the latest and greatest .NET version, and take advantage of newer C# language features too, like async+await. It didn't seem worth it to only go to .NET 4.5 when the point release is available too. 

I guess its probably the `ToolsVersion` stuff that's causing VS2012 to bork: MSBuild changed quite drastically with VS2013 (for the better). There might be an update to VS2012 available that will fix it; the point of the change in VS2013 was to break MSBuild and VS apart so they could be shipped separately.

philpursglove:
I'm on VS2013

alastairs:
> There might be an update to VS2012 available that will fix it

E.g., [this one](http://www.microsoft.com/en-us/download/details.aspx?id=40772)?

adrianbanks:
> There might be an update to VS2012 available that will fix it
> E.g., this one?

That's what I've just installed and it still won't play ball.

adrianbanks:
Just upgraded to VS 2013, so this is ok for me now.

PR #218, opened 2014-03-25
Switch from EntityFramework to Simple Data

adrianbanks:
This pull request removes all trace of EntityFramework and switches the data access to Simple Data. Following this change, a few things should happen:
- we need to figure out how schema changes will be made, as the EF migrations are now gone
- the code could do with a possible bit of refactoring, and definitely shuffling around, as the structure is not right


adrianbanks:
I've rebased this onto `staging` to solve the merge issues.

alastairs:
Ah, I was going to pull the breaking change out of staging because it's failing the deployment...

adrianbanks:
Let me know when you've done that and I'll rebase it again.

alastairs:
I've just done this.

adrianbanks:
Ok, rebased it onto production but it won't build in Team City now due to an invalid failure condition configured on the build (Fail build if number of tests is less by at least 1 compared to the last successful build).

alastairs:
That is a valid failure condition: if the number of tests decreases from the previous build, the build should fail. :smile:

The number of tests in the build of this PR is actually 0 because it failed to compile:

> 2>C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.CSharp.targets(316,9): error MSB3105: The item "GlimpseSecurityPolicy.cs" was specified more than once in the "Sources" parameter.  Duplicate items are not supported by the "Sources" parameter. [C:\TeamCity\buildAgent\work\77c9f48fc195cd03\DDDEastAnglia\DDDEastAnglia.csproj]

I'll try to see if there's a way to surface this information in TeamCity from psake when I get a moment.

adrianbanks:
> That is a valid failure condition: if the number of tests decreases from the previous build, the build should fail.

Not if you've refactored out some code and reduced the number of tests ;)

I've fixed that error - weird that it built locally on my machine.

PR #224, opened 2014-04-14
Clear out the existing sponsors

adrianbanks:
This was missed in the update for 2014 branch already in staging.


alastairs:
Weird, I thought the sponsors' page and the sidebar shared the same code. Evidently not :smile:

PR #227, opened 2014-04-29
Hide voting stats in the admin section when voting hasn't opened

adrianbanks:


DDDEastAnglia.Tests/Admin/AdminHomeControllerTests.cs
```diff
@@ -0,0 +1,79 @@
...
+            var conference = Substitute.For<IConference>();
+            conference.CanSubmit().Returns(true);
+
+            var model = GetViewModel(conference);
+
+            Assert.That(model.ShowVotingStatsLink, Is.False);
+        }
+
+        [Test]
+        public void VotingLink_ShouldBeHidden_WhenVotingIsOpen()
```
alastairs: This test, and the remaining ones in this file, state the Voting link should be hidden when the test asserts that it should be shown.

DDDEastAnglia/Areas/Admin/Controllers/AdminHomeController.cs
```diff
@@ -1,14 +1,30 @@
...
 
 namespace DDDEastAnglia.Areas.Admin.Controllers
 {
     [Authorize(Roles = "Administrator")]
     public class AdminHomeController : Controller
     {
-        // GET: /Admin/AdminHome/
+        private readonly IConferenceLoader conferenceLoader;
+
+        public AdminHomeController(IConferenceLoader conferenceLoader)
```
alastairs, line 13: Interesting that this takes an `IConferenceLoader` when it seems to depend only on the loaded conference itself. What am I missing?

adrianbanks, line 13: The conference needs to be loaded from the database. It can be passed into the constructor, but that would require setting a mapping in Ninject to map the `IConference` interface to a method call on the loader. This doesn't seem correct (especially as we have talked about hosting multiple conferences in the same database, so will need some context to get hold of the correct conference), so instead I'm passing the loader into the controller.

PR #228, opened 2014-04-29
Remove links to submission if session submission is closed

adrianbanks:


DDDEastAnglia.Tests/Controllers/HomeControllerTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        [Test]
+        public void SubmitSessionLinks_ShouldNotBeShown_WhenSessionSubmissionIsNotOpen()
+        {
+            var conference = Substitute.For<IConference>();
+            conference.CanSubmit().Returns(false);
+
+            var conferenceLoader = Substitute.For<IConferenceLoader>();
+            conferenceLoader.LoadConference().Returns(conference);
+            var controller = new HomeController(conferenceLoader);
```
alastairs, line 22: It's worth pulling this construction logic out of the tests and into a method (this is the [SUT Factory](http://blog.ploeh.dk/2009/02/13/SUTFactory/) test pattern).

DDDEastAnglia/Controllers/HomeController.cs
```diff
@@ -1,9 +1,24 @@
...
+using DDDEastAnglia.DataAccess;
+using DDDEastAnglia.Models;
 
 namespace DDDEastAnglia.Controllers
 {
     public class HomeController : Controller
     {
+        private readonly IConferenceLoader conferenceLoader;
+
+        public HomeController(IConferenceLoader conferenceLoader)
```
alastairs, line 12: Again, interesting that this takes an `IConferenceLoader` rather than a conference.

DDDEastAnglia.Tests/Controllers/HomeControllerTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using NSubstitute;
+using NUnit.Framework;
+
+namespace DDDEastAnglia.Tests.Controllers
+{
+    [TestFixture]
+    public sealed class HomeControllerTests
+    {
+        [Test]
+        public void SubmitSessionLinks_ShouldNotBeShown_WhenSessionSubmissionIsNotOpen()
```
alastairs: Might be worth renaming this to `..._ShouldBeHidden_...`, so it's more explicit what the difference is between the tests.

DDDEastAnglia/Views/Home/About.cshtml
```diff
@@ -31,7 +33,18 @@
         </ol>
 
         <p>
-            <a href="#sessions">Sessions</a> are @Html.ActionLink("submitted", "Create", "Session") by members of the community and selected
+            <a href="#sessions">Sessions</a> are 
+            
+            @if (Model.ShowSessionSubmissionLink)
```
alastairs: We have this kind of thing in a few places now: `if (some condition) { render a link } else { render text }`. A new HtmlHelper would be useful for both this situation and the similar situation of `if (some condition) { render a link } else { // do nothing }`.

DDDEastAnglia/Views/Home/About.cshtml
```diff
@@ -31,7 +33,18 @@
...
+            {
+                Html.ActionLink("submitted", "Create", "Session");
+            }
+            else
+            {
+                <text>submitted</text>
+            }
+            
+            by members of the community and selected
```
alastairs: Minor point: the text wrapping is a bit funny here, and on line 67 below.

DDDEastAnglia.Tests/Controllers/HomeControllerTests.cs
adrianbanks, line 22: Yeah, I did that for [this other pull request](https://github.com/dddeastanglia/DDDEastAnglia/pull/227). Once I wrote this code here, I though it would be good to do some refactoring once both are merged to introduce a builder to make a conference loader that returns a populated conference. I left that out of this pull request as I need both pull requests to extract the common functionality required.

DDDEastAnglia/Controllers/HomeController.cs
adrianbanks, line 12: See my comments on the [other pull request](https://github.com/dddeastanglia/DDDEastAnglia/pull/227).

DDDEastAnglia/App_Code/HTMLExtensions.cs
```diff
@@ -41,4 +42,9 @@ public static MvcHtmlString TweetButton(this HtmlHelper htmlHelper, string tweet
                 @"<a href=""https://twitter.com/intent/tweet?text={0}"" class=""btn btn-mini"" target=""_blank""><i class=""icon-twitter""></i> Tweet</a>",
                 encodedTweetText));
     }
+
+    public static MvcHtmlString ConditionalLink(this HtmlHelper htmlHelper, string linkText, string actionName, string controllerName, bool condition)
```
alastairs, line 46: A minor thing: I think I might have called this `ActionLink` and the boolean parameter `displayCondition` or something like that. That way, it highlights the relationship to the provided `ActionLink` helper, but acts as an overload to that method.

alastairs:
The last two commits haven't built on TeamCity, sadly, but I'm going to assume that this is good to merge and see what happens.

Is there _any_ possibility of pushing production forwards to avoid this happening to more branches? Perhaps by pulling my 2014-updates branch for now and merging it again closer to DDDSW?

DDDEastAnglia.Tests/Controllers/HomeControllerTests.cs
alastairs, line 22: That's fair. Feel free to take [Bob](http://nuget.org/packages/BobTheBuilder) for a spin :wink:

DDDEastAnglia/App_Code/HTMLExtensions.cs
adrianbanks, line 46: I think that is a good improvement. I'll pull this out of staging and make the change.

adrianbanks:
> Is there any possibility of pushing production forwards to avoid this happening to more branches? Perhaps by pulling my 2014-updates branch for now and merging it again closer to DDDSW?

I thought production didn't auto-deploy. If so, we can move production forwards but not deploy it until nearer the date.

DDDEastAnglia/App_Code/HTMLExtensions.cs
alastairs, line 46: Ok, could you fix the conflicts while you're there, then? It was the block of commented-out HTML near the top of About.cshtml.

adrianbanks, line 46: Ok, will do.

alastairs:
Fair point. That would be a break with our current process, though, which is to push production forward only when we deploy so that the two are always in sync.

adrianbanks:
True, but I think it is of benefit to do this at the moment. Ideally, the site would go into "maintenance mode" until the announcement date, but we don't have that feature.

alastairs:
Good point. That might be a useful feature to implement post DDDEA 2014...

PR #236, opened 2014-04-30
Add the initial sponsors

adrianbanks:
My changes are in the last commit only.


alastairs:
Please could you add Red Gate to this pull request as well. They are signed up to sponsor us too.

adrianbanks:
I've only added the sponsors that we have already received the money from.

alastairs:
Yes, but as previously discussed (and agreed) Red Gate won't be paying us until much nearer the event.

PR #240, opened 2014-05-10
Lock down pages

adrianbanks:
On several pages, you could edit things belonging to other users. I think I've caught them all, but I'm not 100% sure.

### DIFF
--- a/DDDEastAnglia.Tests/Controllers/SessionControllerTests.cs
+++ b/DDDEastAnglia.Tests/Controllers/SessionControllerTests.cs
@@ -3,6 +3,7 @@ using DDDEastAnglia.Controllers;
 using DDDEastAnglia.DataAccess;
 using DDDEastAnglia.Domain;
 using DDDEastAnglia.Helpers;
+using DDDEastAnglia.Models;
 using NSubstitute;
 using NUnit.Framework;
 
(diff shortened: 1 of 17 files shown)


alastairs:
Please could you explain a bit more about how this change fixes the issues identified? The `UserNameFilter`, for example, at first glance, doesn't seem to be important to the solution.

adrianbanks:
The `UserNameFilter` is purely a mechanism that allows the testing of the controller without the need to rely on `HttpContext` (taken from a Stackoverflow post that you even commented on :wink:). The filter passes the username of the current user into the action as a parameter without it needing to be specified on the url.

alastairs:
> taken from a Stackoverflow post that you even commented on :wink:

Bloody hell, really?! :smile: Ok. I'll take a closer look later.

adrianbanks:
Yes -> http://stackoverflow.com/a/1389798/116923

DDDEastAnglia.Tests/Controllers/SessionControllerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+    public sealed class SessionControllerTests
+    {
+        [Test]
+        public void CannotSeeSessionList_WhenTheConferenceSaysThatSessionsCannotBeShown()
+        {
+            var conference = Substitute.For<IConference>();
+            conference.CanShowSessions().Returns(false);
+            var conferenceLoader = Substitute.For<IConferenceLoader>();
+            conferenceLoader.LoadConference().Returns(conference);
+            var controller = new SessionController(conferenceLoader, Substitute.For<IUserProfileRepository>(), Substitute.For<ISessionRepository>(), Substitute.For<ISessionSorter>());
```
alastairs: If you wish, you can use a [SUT Factory](http://blog.ploeh.dk/2009/02/13/SUTFactory/) to decouple the tests from the constructor of the system under test. 

It also looks like the `SessionController` has a few too many dependencies. I wonder if we can refactor them a bit? The session repository and sorter look related, for example.

Anyway, neither of these items are blocking the merge of this pull request.

alastairs: >  neither of these items are blocking the merge of this pull request.

That said, looking over the tests in a bit more detail, I see two fixtures here: one for when manipulating sessions is forbidden by the stage of the process the conference has reached (e.g., submission isn't open yet, submissions have closed, etc.), and one for manipulating other people's sessions. With two fixtures, there would be one SUT Factory per fixture, and it would refactor quite nicely. If you have time to do this, I think it would be a useful change.

DDDEastAnglia/Controllers/BannerController.cs
```diff
@@ -1,20 +1,26 @@
...
+            if (bannerQuery == null)
+            {
+                throw new ArgumentNullException("bannerQuery");
+            }
+            
+            this.bannerQuery = bannerQuery;
         }
 
-        public ActionResult Details(string id)
+        public ActionResult Details()
```
alastairs, line 21: Why has the `id` parameter been removed?

DDDEastAnglia/Controllers/UserNameFilter.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace DDDEastAnglia.Controllers
+{
+    public sealed class UserNameFilter : ActionFilterAttribute
+    {
+        private readonly string userNameParameterName;
+
+        public UserNameFilter(string userNameParameterName)
+        {
+            this.userNameParameterName = userNameParameterName;
```
alastairs: Should this parameter be null/empty checked before use? Given it's an attribute, it's a bit questionable, but my instinct says "better to be safe" :smile:

DDDEastAnglia/Controllers/UserNameFilter.cs
```diff
@@ -0,0 +1,27 @@
+using System.Web.Mvc;
+
+namespace DDDEastAnglia.Controllers
+{
+    public sealed class UserNameFilter : ActionFilterAttribute
```
alastairs: Convention says attribute classes are named `FooAttribute` (like the `ActionFilterAttribute` you're inheriting from).

alastairs:
Overall, this looks like a good change. Should be good to merge after the questions have been answered.

DDDEastAnglia/Controllers/BannerController.cs
adrianbanks, line 21: The id is the short name of the conference. To load the banner, the short name was hard-coded into the `index.cshtml` file. I've taken it out as the `BannerModelQuery` now uses the `ConferenceLoader` to get the current conference. The logic for what is the "current conference" is contained solely within the conference loader.

DDDEastAnglia/Controllers/UserNameFilter.cs
adrianbanks: Yes, probably.

DDDEastAnglia/Controllers/UserNameFilter.cs
adrianbanks: Yes, I agree.

adrianbanks:
> It also looks like the SessionController has a few too many dependencies. I wonder if we can refactor them a bit? The session repository and sorter look related, for example.

I thought the same when changing it, but didn't have the time to do much about it. Something for later.

DDDEastAnglia/Controllers/BannerController.cs
alastairs, line 21: > The logic for what is the "current conference" is contained solely within the conference loader.

:+1: :sunny: :sunglasses: :cocktail:

DDDEastAnglia.Tests/Controllers/SessionControllerTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            var session = new Session { SessionId = 1, SpeakerUserName = "bob" };
+            var controller = SessionControllerFactory.Create(session);
+
+            var result = controller.DeleteConfirmed("fred", 1);
+
+            Assert.That(result, Is.InstanceOf<HttpUnauthorizedResult>());
+        }
+
+        private static class SessionControllerFactory
```
alastairs, line 83: The two methods in this class represent the division in the test fixture: some tests use one, the rest use the other. I'm happy to merge as-is, just thought I'd point it out for future reference.

adrianbanks, line 83: Yes I know. The only reason I didn't split them was due to lack of inspiration for sensible names :)

adrianbanks:
> You're checking for null twice in this. I normally treat null strings and empty strings as equivalent, because the string.Empty is like a Null Object implementation for string.

I've swayed back and forth on this a few times now. Semantically they are different, especially in the types of exceptions thrown - it is a useful diagnostic to differentiate the two. However, an invalid string is an invalid string, so it the same.

alastairs:
> Semantically they are different, especially in the types of exceptions thrown

If it makes you more comfortable, an `ArgumentException` would be appropriate to both types of exception :smile: 

> it is a useful diagnostic to differentiate the two. However, an invalid string is an invalid string, so it the same.

I'm not sure it is necessarily, because it's invalid as you say, and it's invalid in the same way - no content in the string - rather than being invalid in a different way (e.g., `"foo"` for a postcode string). I _think_ we're quite good at treating `null` and empty strings as the same in most places in this code base, so I'd prefer not to introduce a difference here.

PR #242, opened 2014-05-12
Update the conference code for the 2014 event

adrianbanks:
_Once this is merged, an upgrade script will need to be run on the database._

I have this script so can run it, but don't want to run it until this code is merged.


alastairs:
Please run the script on the staging database.

PR #243, opened 2014-05-12
Live dates from the database

adrianbanks:
The change shows the live dates from the database in the side bar and header instead of hard-coded ones.

### DIFF
--- a/DDDEastAnglia/Controllers/TimelineController.cs
+++ b/DDDEastAnglia/Controllers/TimelineController.cs
@@ -9,6 +9,7 @@ namespace DDDEastAnglia.Controllers
     public class TimelineController : Controller
     {
         private const string DateOnlyPattern = "dddd d MMMM yyyy";
+        private const string TimeOnlyPattern = "H:mm";
         private const string DateAndTimePattern = "dddd d MMMM yyyy, H:mm";
 
         private readonly ICalendarItemRepository calendarItemRepository;
@@ -21,6 +22,22 @@ namespace DDDEastAnglia.Controllers
             }
             
             this.calendarItemRepository = calendarItemRepository;
+        }
+
+        public ActionResult ConferenceDate()
+        {
+            var conference = calendarItemRepository.GetFromType(CalendarEntryType.Conference);
+            string conferenceDate = conference.StartDate.ToString(DateOnlyPattern);
+            return PartialView("_ConferenceDateTime", conferenceDate);
+        }
+
+        public ActionResult ConferenceTime()
+        {
+            var conference = calendarItemRepository.GetFromType(CalendarEntryType.Conference);
+            string startTime = conference.StartDate.ToString(TimeOnlyPattern);
+            string endTime = conference.EndDate.Value.ToString(TimeOnlyPattern);
+            string conferenceTimes = string.Format("{0} to {1}", startTime, endTime);
+            return PartialView("_ConferenceDateTime", conferenceTimes);
         }
 
         public ActionResult Details()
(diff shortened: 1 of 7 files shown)


DDDEastAnglia/Controllers/TimelineController.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Web.Mvc;
+using DDDEastAnglia.DataAccess;
+using DDDEastAnglia.Domain.Calendar;
+using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.Controllers
+{
+    public class TimelineController : Controller
+    {
+        private const string DateOnlyPattern = "dddd d MMMM yyyy";
```
alastairs, line 11: Do we want to use the Culture-specific [`ToLongDateString()`](http://msdn.microsoft.com/en-us/library/system.datetime.tolongdatestring%28v=vs.110%29.aspx) method and friends instead?

DDDEastAnglia/Views/Timeline/_ConferenceDateTime.cshtml
```diff
@@ -0,0 +1,3 @@
+﻿@model string
+
+@Model
```
alastairs: This seems a bit nuts. Can't you just return a `ContentResult` instead?

DDDEastAnglia/Controllers/TimelineController.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (endDateTime == null)
+            {
+                return string.Empty;
+            }
+
+            var endDateTimeValue = endDateTime.Value;
+            var timeOfDay = endDateTimeValue.TimeOfDay;
+
+            // if the end time is midnight, we need to subtract a little to push it back to the previous date
+            if (timeOfDay.Hours == 0 && timeOfDay.Minutes == 0 && timeOfDay.Seconds == 0)
```
alastairs, line 82: Ouch. This looks like it's ripe for timezone issues... Or does the fact that we're dealing with `DateTimeOffset`s rather than `DateTime`s fix those issues for us?

alastairs:
:+1: to this change in principle, and it's nearly there.

DDDEastAnglia/Controllers/TimelineController.cs
adrianbanks, line 11: `ToLongDateString()` doesn't include the day in the en-GB culture (none of the built-in formats include the day name in en-GB). I think it very useful to show the day.

I could use `ToShortTimeString()` for the time, but in en-GB it includes the leading `0`, which looks a bit odd in the sidebar.

DDDEastAnglia/Views/Timeline/_ConferenceDateTime.cshtml
adrianbanks: Yeah - not sure what I was thinking when I did this!

DDDEastAnglia/Controllers/TimelineController.cs
adrianbanks, line 82: `DateTimeOffset` is always relative to UTC, so it shouldn't suffer from that problem.

This was the best approach I could come up with to closing at midnight but showing the previous date in the UI. Any ideas?

alastairs, line 82: No further ideas :smile: If you're happy this won't be open to the timezone issues we had last year (not that we will be crossing timezones this year, of course) then I'm happy to merge this.

DDDEastAnglia/Controllers/TimelineController.cs
alastairs, line 11: Okey-dokey.

DDDEastAnglia/Controllers/TimelineController.cs
adrianbanks, line 82: Yes, I'm (fairly) confident we won't have a problem.

Issue #246, opened 2014-05-13
Links in the admin section are broken


adrianbanks:
> The model item passed into the dictionary is of type 'System.Collections.Generic.List`1[DDDEastAnglia.Areas.Admin.Models.UserModel]', but this dictionary requires a model item of type 'DDDEastAnglia.Areas.Admin.Models.MenuViewModel'.

```
[InvalidOperationException: The model item passed into the dictionary is of type 'System.Collections.Generic.List`1[DDDEastAnglia.Areas.Admin.Models.UserModel]', but this dictionary requires a model item of type 'DDDEastAnglia.Areas.Admin.Models.MenuViewModel'.]
   System.Web.Mvc.ViewDataDictionary`1.SetModel(Object value) +378
   System.Web.Mvc.ViewDataDictionary.set_Model(Object value) +47
   System.Web.Mvc.ViewDataDictionary..ctor(ViewDataDictionary dictionary) +614
   System.Web.Mvc.ViewDataDictionary`1..ctor(ViewDataDictionary viewDataDictionary) +37
   System.Web.Mvc.WebViewPage`1.SetViewData(ViewDataDictionary viewData) +98
   System.Web.Mvc.WebViewPage.set_ViewData(ViewDataDictionary value) +38
   System.Web.Mvc.WebViewPage.ConfigurePage(WebPageBase parentPage) +267
   System.Web.WebPages.<>c__DisplayClass7.<RenderPageCore>b__6(TextWriter writer) +218
   System.Web.WebPages.HelperResult.WriteTo(TextWriter writer) +10
   System.Web.WebPages.WebPageBase.Write(HelperResult result) +71
   System.Web.WebPages.WebPageBase.RenderSurrounding(String partialViewName, Action`1 body) +64
   System.Web.WebPages.WebPageBase.PopContext() +246
   System.Web.WebPages.WebPageBase.ExecutePageHierarchy(WebPageContext pageContext, TextWriter writer, WebPageRenderingBase startPage) +95
   System.Web.Mvc.RazorView.RenderView(ViewContext viewContext, TextWriter writer, Object instance) +743
   System.Web.Mvc.BuildManagerCompiledView.Render(ViewContext viewContext, TextWriter writer) +382
   System.Web.Mvc.ViewResultBase.ExecuteResult(ControllerContext context) +431
   System.Web.Mvc.ControllerActionInvoker.InvokeActionResult(ControllerContext controllerContext, ActionResult actionResult) +39
[...]

PR #247, opened 2014-05-13
Fix issue #246

adrianbanks:
Extract the banner links out into a separate view


DDDEastAnglia/Areas/Admin/Controllers/AdminHomeController.cs
```diff
@@ -22,9 +22,22 @@ public AdminHomeController(IConferenceLoader conferenceLoader)
 
         public ActionResult Index()
         {
+            var menuViewModel = CreateMenuViewModel();
+            return View(menuViewModel);
+        }
+
+        public ActionResult BannerLinks()
```
alastairs: Please could you rename this to NavLinks (assuming this is what you mean)? I keep thinking this is related to the notification/announcement banner, rather than the nav bar.

(Ditto the other changes, such as the View below.)

alastairs:
Are there already tests around this functionality?

adrianbanks:
> this functionality

What are you referring to? There are tests to check the voting link is hidden when appropriate.

PR #248, opened 2014-05-13
Remove the broken version info text

adrianbanks:
It doesn't work as we don't deploy a git repository any more.


alastairs:
> It doesn't work as we don't deploy a git repository any more.

Did it ever work? We've always deployed to staging using the GitHub's webhook (so far as I can remember, at least), and production is deployed as a `git push`. I can't think off the top of my head in what circumstances it would have worked.

adrianbanks:
I definitely saw it working somewhere, but not for a long time. Seems pointless keeping it in now.

PR #252, opened 2014-05-20
Fix #251 Prevent the agenda being seen unless it should be

### LINKED ISSUE #251: Last year's agenda is still available to view.
Is this something we missed, or left up on purpose? http://dddeastanglia.com/Agenda

adrianbanks:
This also restricts access to the registration and accommodation page unless registration is open.


alastairs:
Looks good - a nice change.

PR #254, opened 2014-05-22
Add CRedit360 logo.

alastairs:
CRedit360 have confirmed their sponsorship of DDDEA and have been invoiced.


adrianbanks:
Might be worth shrinking the logo a touch to add some white space around its edge, otherwise it may look rather large next to the other logos on the page. Let's see how it looks first.

adrianbanks:
Doesn't look as big as it appeared. It's fine as it is.

PR #261, opened 2014-06-02
Fix #260 YSOD when attempting to reset a forgotten password

### LINKED ISSUE #260: Yellow Screen of Death if you click forgotton password link
Url changes to http://www.dddeastanglia.com/ResetPassword/Start but all you get is a YSOD

adrianbanks:
Put in the default constructor that got accidentally taken out.

### DIFF
--- a/DDDEastAnglia/Controllers/ResetPasswordController.cs
+++ b/DDDEastAnglia/Controllers/ResetPasswordController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Web.Mvc;
 using DDDEastAnglia.DataAccess;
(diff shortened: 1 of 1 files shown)


alastairs:
What happened in this pull request?? Infinite commits, 0 changed files...

adrianbanks:
I merged it manually to the head of production, moved production forward to this fix, then rebased staging onto the new production. I suspect something odd happened to the diffs in the process. [This](https://github.com/dddeastanglia/DDDEastAnglia/commit/8dc7759283801b98b54df4d3c958e543174bebeb) is the only change I made, which was to put back in a line of code that got taken out by mistake when we moved to Simple.Data.

PR #263, opened 2014-06-02
Remove a hangover from before we had Ninject wired in

adrianbanks:
I'm a bit confused by this. I can only think that I don't understand how Ninject's convention-based binding works.

I appear to need a binding for:

`IMailHostSettingsProvider` -> `SmtpHostSettingsProvider`

but not for:

`IMessageFactory` - > `SendGridMessageFactory`
`IEmailSender` -> `SendGridEmailSender`
`IAppSettingsProvider` -> `WebConfigurationAppSettingsProvider`

I don't understand why the difference.


alastairs:
I've had a look and can't figure it out as yet either :unamused:

adrianbanks:
This change does work, but I don't really understand the difference in the configuration.

PR #265, opened 2014-06-04
Fix #245 Add a preview and closed date to the site

### LINKED ISSUE #245: Add new dates to the database
We should add two addition dates to the database:
- conference hibernation
- conference announcement

Conference hibernation will put the site into minimal mode, showing only basic information about the previous conference.

Conference announcement is a date where the site will automatically come alive again to reveal a new conference.

adrianbanks:
There are now two extra dates available for a conference - a preview date and a closed date.

In the preview date range, all urls to the site will get redirected to a `/preview` page that shows a holding/teaser page (e.g. coming soon...).

In the closed date range, all urls to the site will get redirected to a `/closed` page that shows that the conference is over for the year.


DDDEastAnglia/App_Start/Filters/ConferenceIsClosedFilterAttribute.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Web.Mvc;
+using DDDEastAnglia.DataAccess;
+
+namespace DDDEastAnglia.Filters
+{
+    public class ConferenceIsClosedFilterAttribute : ActionFilterAttribute
+    {
+        private readonly IConferenceLoader conferenceLoader;
+
+        public ConferenceIsClosedFilterAttribute(IConferenceLoader conferenceLoader)
```
alastairs: This needs a null check.

DDDEastAnglia/App_Start/Filters/ConferenceIsInPreviewFilterAttribute.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Web.Mvc;
+using DDDEastAnglia.DataAccess;
+
+namespace DDDEastAnglia.Filters
+{
+    public class ConferenceIsInPreviewFilterAttribute : ActionFilterAttribute
+    {
+        private readonly IConferenceLoader conferenceLoader;
+
+        public ConferenceIsInPreviewFilterAttribute(IConferenceLoader conferenceLoader)
```
alastairs, line 11: This needs a null check too.

adrianbanks:
Updated.

DDDEastAnglia/Domain/Conference.cs
```diff
@@ -29,32 +24,47 @@ public int Id
...
+        {
+            return GetCalendarEntry(CalendarEntryType.Preview).IsOpen();
+        }
+
+        public bool IsClosed()
+        {
+            return GetCalendarEntry(CalendarEntryType.Closed).IsOpen();
+        }
+
+        private bool IsNotInSpecialMode()
```
alastairs: This might be a matter of style, but I prefer methods like this to return a positive value, rather than a negative one. I know the result is not current being negated, but I think it would be more readable to invert this logic and use `!` on each usage instead. Or perhaps rename the method to be something like `ConferenceTimelineIsActive()`.

DDDEastAnglia/Views/Home/Preview.cshtml
```diff
@@ -0,0 +1,24 @@
+﻿@{
+    Layout = null;
+}
+
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        @Html.Partial("_HeadSection")
```
alastairs, line 8: I think this kind of thing (and the footer below) is what Razor's `@section` feature is for. Partials are more for reusable controls, with behaviour. Just a thought.

DDDEastAnglia/Domain/Conference.cs
adrianbanks: I'll change this. I wasn't happy with the name, but couldn't come up with something better at the time.

DDDEastAnglia/App_Start/Filters/ConferenceIsInPreviewFilterAttribute.cs
alastairs, line 11: This attribute doesn't seem to be used anywhere??

DDDEastAnglia/Views/Home/Preview.cshtml
adrianbanks, line 8: I think there was a reason that I couldn't use sections. I'll have a play tonight and see if I can remember.

DDDEastAnglia/App_Start/Filters/ConferenceIsInPreviewFilterAttribute.cs
adrianbanks, line 11: It's not actually an attribute in the normal sense - it is applied somewhere else. The entry point is via `FilterConfig`. I think I originally had it called `ConferenceIsInPreviewFilter`, as can be seen by the naming of the tests. I might name it back that way to make it less confusing.

alastairs, line 11: Umm, it's still called `ConferenceIsInPreviewFilterAttribute`...

adrianbanks, line 11: Typo ;) Try reading it again!

DDDEastAnglia/Views/Home/Preview.cshtml
adrianbanks, line 8: Ok, figured out why a section isn't valid here. Sections are used in a layout to render part of the page. Each page that consumes the layout can then insert content. For these pages, they don't use the standard default layout as they don't need any of the components of the page (i.e. navbar, sidebar, header logo, etc...). In effect, these pages (closed & preview) are more like the layout than a page that consumes.

I do however want to share the header and the footer so that I don't copy/paste the whole lot, so a partial is the way to go.

adrianbanks:
The attributes aren't used anywhere as I want them to apply to _everything_ in the site. To avoid having to apply them to everything (and also to remember to put them on anything new), I've inverted their behaviour so that they apply by default and there are two new attributes that exempt certain actions from the behaviour. It's a bit of an odd way of doing it, but from lots of searching it seems to be the only reasonable approach.

PR #266, opened 2014-06-05
Re-introduce database migrations

adrianbanks:
We lost the ability to do schema changes when we removed EF (although in reality we could never do them easily). This re-introduces the ability using [Fluent Migrator](https://github.com/schambers/fluentmigrator/wiki).

When this gets pushed to `staging` and `production`, a manual database upgrade will need to be run to get the database into the correct initial state. The `ManualUpgrade.sql` is a script for this, but it might need tweaking depending on the actual names of the auto-generated key names.

Once the manual upgrade has run, further schema changes can be made by simply adding a new migration to the `DDDEastAnglia.DatabaseMigrations` project and it will get automatically applied next time the website starts.


DDDEastAnglia.DatabaseMigrations/Migrator.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                throw new ArgumentNullException("connectionString");
+            }
+            
+            this.connectionString = connectionString;
+        }
+
+        public void MigrateToLatestSchema()
+        {
+            var todaysDate = DateTime.Today.ToString("yyyyMMdd");
```
alastairs, line 25: Do you expect anyone to ever set a schema version to use a future date?

DDDEastAnglia.DatabaseMigrations/Migrator.cs
```diff
@@ -0,0 +1,41 @@
...
+            var todaysSchemaVersion = long.Parse(todaysDate);
+            MigrateTo(todaysSchemaVersion);
+        }
+
+        public void MigrateTo(long targetVersion)
+        {
+            var options = new MigrationOptions { PreviewOnly = false,  Timeout = 60 };
+            var announcer = new NullAnnouncer();
```
alastairs, line 33: The announcer sounds like it would provide cool functionality for the Admin part of the site: perhaps displaying the progress of the current migration, or simply reporting success/failure of various migrations. Is this the kind of thing it's designed for, or have I misunderstood?

DDDEastAnglia.DatabaseMigrations/Migrator.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public void MigrateToLatestSchema()
+        {
+            var todaysDate = DateTime.Today.ToString("yyyyMMdd");
+            var todaysSchemaVersion = long.Parse(todaysDate);
+            MigrateTo(todaysSchemaVersion);
+        }
+
+        public void MigrateTo(long targetVersion)
```
alastairs: This doesn't seem to warrant being made public yet: we only ever migrate to the latest schema. Unless you have plans to add in-app roll back or similar?

alastairs:
This looks like a really nice change; the migrations are fantastically readable! Are the migrations testable at all? For example, does Fluent Migrator provide a no-op `ProcessorFactory`, `MigrationContext` or `MigrationRunner` (whatever the appropriate extension point is) that allows you to verify the actions of the migration in a test without hitting a database for reals?

What happens if we write a migration script that fails for some reason? Each time the application starts, it will try to migrate to (or through) that version and fail. Will the application fail to start? Will it YSoD? Will it migrate to an earlier version and result in bugs throughout the site?

DDDEastAnglia.DatabaseMigrations/Migrator.cs
adrianbanks, line 25: Not really, no. It was either hard code the current schema somewhere which would require the person making a schema change to remember to change it also, or just assume that we would be sensible with the dates and use the current date for the latest schema (the schema version is actually just a number).

DDDEastAnglia.DatabaseMigrations/Migrator.cs
adrianbanks, line 33: Yes, that is what it is for, although the ones that come out of the box don't give us anything that useful:
- `NullAnnouncer` - does no logging
- `ConsoleAnnouncer` - logs to `Console`
- `TextWriterAnnouncer` - logs to a `TextWriter`

We could extend this (perhaps using `TextWriterAnnouncer`) to capture a log and put it in the admin section, but at the moment I don't think it worth the effort. From experience in developing this, if the schema upgrade fails you get an exception at application start up. Next time the application starts, you get the same exception as it re-attempts to upgrade the schema. This exception has been useful every time I got it.

DDDEastAnglia.DatabaseMigrations/Migrator.cs
adrianbanks: I didn't plan on doing that, so I'll make it private.

adrianbanks:
> Are the migrations testable at all?

I gave that some thought and didn't really know what I would test without doing a real upgrade on a database (but the problem then is doing the comparison of the reference database and the upgraded one - at least without using a commercial tool ;)).

> What happens if we write a migration script that fails for some reason? Each time the application starts, it will try to migrate to (or through) that version and fail. Will the application fail to start? Will it YSoD?

Yes, it will attempt to get to the correct schema every time, and fail, and YSOD (we need a [nicer error page](https://github.com/dddeastanglia/DDDEastAnglia/issues/267)).

> Will it migrate to an earlier version and result in bugs throughout the site?

I think it will migrate to the latest version it can and stop there, so yes the database schema may not match the code. In these cases, I would hope (at least for now) that we will monitor the upgrades when we push to staging/production, so I don't see that as a problem. If it goes wrong, we can sort it out manually. We could try to put the management of this in a UI somewhere, but it would have to be out of the context of the rest of the application as everything that is present at the moment relies on a correct database schema.

alastairs:
> > Are the migrations testable at all?
> 
> I gave that some thought and didn't really know what I would test without doing a real upgrade on a database (but the problem then is doing the comparison of the reference database and the upgraded one - at least without using a commercial tool ;)).

Oh, I'd hoped that FluentMigrator would provide a way to run the migrations in-memory only, or something like that - a mocked database, I guess. The syntax is about as close to English and/or SQL as can be desired, though, and there's no complicated logic in the migrations yet, so I'm happy to merge as-is.

adrianbanks:
> Oh, I'd hoped that FluentMigrator would provide a way to run the migrations in-memory only, or something like that - a mocked database, I guess.

What would you assert without just duplicating the syntax of the test action?

PR #269, opened 2014-06-08
Fix #162 Record the position in the list of each voted session

### LINKED ISSUE #162: Record the position of the voted for session in the random ordered list

adrianbanks:
This will require a schema change:

```
ALTER TABLE [dbo].[Votes] ADD [PositionInList] [int] NOT NULL
```

As a result, it will need some level of integration with #266, as that contains a script to create the default structure.


alastairs:
#266 is now merged.

adrianbanks:
I've rebased this onto the database migrations branch, so it should merge more easily.

alastairs:
Let's see how well the migrations work...

PR #270, opened 2014-06-09
Fix #268 User pages in the admin section

### LINKED ISSUE #268: Links in the User section in the Admin area
The links on the User Index page (Details/Edit/Delete) all point to the Session controller e.g. http://dddeastanglia.com/Admin/Session/Details/1308 is the Details link for Dermot Kilroy.

adrianbanks:
I've fixed the links, but also implemented:
- viewing user details
- editing of all user details
- deleting a user

### DIFF
--- a/DDDEastAnglia/Areas/Admin/Content/admin.css
+++ b/DDDEastAnglia/Areas/Admin/Content/admin.css
@@ -1,4 +1,8 @@
-﻿.table th {
+﻿body {
+    margin-top: 50px;
+}
+
+.table th {
     text-transform: capitalize;
 }
 
--- a/DDDEastAnglia/Areas/Admin/Content/admin.css
+++ b/DDDEastAnglia/Areas/Admin/Content/admin.css
@@ -6,6 +6,10 @@ a { text-decoration: underline; }
 
 a.btn, .dropdown-menu a, a.brand, a [class^="icon-"] { text-decoration: none; }
 
+label:first-letter {
+    text-transform: capitalize;
+}
+
 .table th {
     text-transform: capitalize;
 }
--- a/DDDEastAnglia/Areas/Admin/Content/admin.css
+++ b/DDDEastAnglia/Areas/Admin/Content/admin.css
@@ -1,16 +1,4 @@
-﻿body {
-    padding-top: 50px;
-}
-
-a { text-decoration: underline; }
-
-a.btn, .dropdown-menu a, a.brand, a [class^="icon-"] { text-decoration: none; }
-
-label:first-letter {
-    text-transform: capitalize;
-}
-
-.table th {
+﻿.table th {
     text-transform: capitalize;
 }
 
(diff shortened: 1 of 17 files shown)


alastairs:
Hmmm, is there any possibility of pulling the two parts of this work out into separate pull requests...?

Why did you need to split Bootstrap out into a separate style bundle? Does this result in a separate CSS URL as well? I can't quite remember how the bundles are "rendered" in production.

adrianbanks:
> Hmmm, is there any possibility of pulling the two parts of this work out into separate pull requests...?

Which two parts are you referring to?

> Why did you need to split Bootstrap out into a separate style bundle? Does this result in a separate CSS URL as well? I can't quite remember how the bundles are "rendered" in production.

Because bootstrap wasn't being included in the admin area layout page, and in the bundle it was bundled with the css files for the main site was there (which we don't use in the admin area). Having said that, I've just looked at the diffs and what I thought I'd done isn't what is here, so I'll tweak it tonight.

alastairs:
I meant this bit:

> I've fixed the links, but also implemented [other stuff]

It would be useful to have the bugfix in one pull request and the new features in a separate one.

adrianbanks:
Didn't think it was worth separating the two, as fixing the links then means that the link urls are correct, but 404 when clicked on as there are no actions to handle them :confused:

alastairs:
Oh! I hadn't appreciated that, sorry. No worries then; just the question about Bootstrap to resolve, and I'll do a quick review of the rest of the change now.

DDDEastAnglia/Areas/Admin/Content/admin.css
```diff
@@ -81,3 +85,42 @@ th.barchart, td.barchart {
...
+    position: relative;
+    padding: 35px 15px 15px 15px;
+    border: 1px solid #DDD;
+    -webkit-border-radius: 4px;
+    -moz-border-radius: 4px;
+    border-radius: 4px;
+    margin-bottom: 20px;
+}
+
+.group-box:after {
```
alastairs: I checked this out locally to see what it looks like... Nice styling!

adrianbanks: Copied directly from the bootstrap site (although oddly not part of bootstrap itself).

DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
```diff
@@ -2,12 +2,13 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
-        <title>DDD East Anglia Admin @ViewBag.Title</title>
+        <title>DDD East Anglia Admin @(string.IsNullOrWhiteSpace(ViewBag.Title) ? "" : "-") @ViewBag.Title</title>
```
alastairs, line 5: Hmmm, this should probably be taken care of as part of the title rather than as part of the view. That said, it is purely presentational, so...

adrianbanks, line 5: I wanted the page title to make sense, so needed some separation otherwise they read "DDD East Anglia Admin Users". As it is needed for every page, this seemed like the best place, rather than adding a leading " - " onto each title in each page.

adrianbanks:
I think I've sorted out the styles now - seems to look ok after the css shuffle.

PR #272, opened 2014-06-09
Put the connection strings into a 'secret' file

adrianbanks:
Similar to `appSettingsSecret.config`, there is now support for a `connectionStringsSecret.config` file to contain the database connection strings. This file should not be checked into git, and should contain machine-specific connection strings. The connection string on the live sites is configured in the Azure portal.

### DIFF
--- a/DDDEastAnglia/DDDEastAnglia.csproj
+++ b/DDDEastAnglia/DDDEastAnglia.csproj
@@ -542,6 +542,10 @@
     <Content Include="Content\font\FontAwesome.otf" />
     <Content Include="Areas\Admin\Views\Voting\DuplicateVotes.cshtml" />
     <Content Include="Areas\Admin\Views\AdminHome\NavigationLinks.cshtml" />
+    <ExcludeFromPackageFiles Include="connectionStringsSecret.config">
+      <FromTarget>Project</FromTarget>
+      <SubType>Designer</SubType>
+    </ExcludeFromPackageFiles>
     <None Include="Scripts\jquery-1.9.1.intellisense.js" />
     <Content Include="Scripts\bootstrap.js" />
     <Content Include="Scripts\html5shiv-printshiv.js" />
--- a/DDDEastAnglia/Web.config
+++ b/DDDEastAnglia/Web.config
@@ -15,9 +15,7 @@
     </sectionGroup>
     <section name="glimpse" type="Glimpse.Core.Configuration.Section, Glimpse.Core" />
   </configSections>
-  <connectionStrings>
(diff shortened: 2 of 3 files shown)


adrianbanks:
Not sure why this isn't building - I've set the properties of the new file to match those of the existing "secret" file.

alastairs:
The build is failing because the file doesn't exist:

> web.config(18): error ASPCONFIG: Unable to open configSource file 'connectionStringsSecret.config'. [DDDEastAnglia\DDDEastAnglia.csproj]

A related issue is that Azure requires the connection string to be defined _somewhere_ before it can be overridden with the settings from the portal. What we can do to sort all of this is to check in the new `connectionStringsSecret.config` file, perhaps with an empty value, and then explicitly ignore this file so git does not track any further changes to the file. This will allow the deployments to succeed (and should fix the build too), while still buying us the benefit of removing the churn on this part of the web.config file (and potential security hazards of people checking in credentials to their development databases).

adrianbanks:
> What we can do to sort all of this is to check in the new connectionStringsSecret.config file, perhaps with an empty value, and then explicitly ignore this file so git does not track any further changes to the file. This will allow the deployments to succeed (and should fix the build too), while still buying us the benefit of removing the churn on this part of the web.config file (and potential security hazards of people checking in credentials to their development databases).

Sounds like a plan. I'll make that change (will have to be two separate pull requests I believe). I was confused however how the other secret file is in the project but doesn't exist in git, yet the deployment still works.

alastairs:
>  I was confused however how the other secret file is in the project but doesn't exist in git, yet the deployment still works.

Simplez: the implementations for AppSettings and ConnectionStrings are different :angry: I'm pretty certain this is a difference baked into .NET years ago rather than a mistake in Azure.

adrianbanks:
I wouldn't have thought that difference was coded into the deployment though. Strange. I'll make the change as we said.

alastairs:
Sorry, lack of clarity on my part. What I meant was "the deployment breaks" in the general sense: it fails at runtime rather than during the deployment.

adrianbanks:
I thought that the deployment is failing because a file is in the project that is not present on disk. If that is the case, the appsettings file is exactly the same, yet somehow that gets through the deployment. That's why I'm a bit confused about it.

alastairs:
> I thought that the deployment is failing because a file is in the project that is not present on disk.

No, deployment hasn't been attempted yet AFAIK. The _build_ is failing for this reason :smile:

adrianbanks:
Ah! That might explain it.

alastairs:
:+1:

PR #273, opened 2014-06-09
Redirect the user to the page they requested after logging in

adrianbanks:
Without this, they end up on the front page.


alastairs:
This was in the ViewBag but not in the form action? Odd.

PR #275, opened 2014-06-11
Sessions pages in admin section

adrianbanks:
This is based on #270 as it shares a lot of the styling changes, so that should be merged first.

Style the session details, edit and delete pages to match the newly-created users pages.

### DIFF
--- a/DDDEastAnglia.Tests/ActionResultExtensions.cs
+++ b/DDDEastAnglia.Tests/ActionResultExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.Web.Mvc;
+﻿using System.Net;
+using System.Web.Mvc;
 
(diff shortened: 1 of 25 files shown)


alastairs:
Where are the tests for the new actions? I think I might have missed this in the user's admin pages too... :frowning:

adrianbanks:
I couldn't see what to test other than the implementation when I looked at it originally. Looking at it again, I think there is some behaviour for some of the actions, but not all. Any suggestions?

alastairs:
Well, off the top of my head, there's the interactions with the repositories to verify: ensuring the controller delegates correctly.

adrianbanks:
> there's the interactions with the repositories to verify

Hmm, feels a bit like testing the implementation. I'll take a look at see what I can do.

PR #276, opened 2014-06-11
User pages in the admin section

adrianbanks:
I've pulled out the merge of #270 so it can be merged with the newly-added tests all in one go.

@alastairs: I'd like an opinion on some of these, as a few of them feel like I'm testing just the implementation.

### DIFF
--- a/DDDEastAnglia.Tests/ActionResultExtensions.cs
+++ b/DDDEastAnglia.Tests/ActionResultExtensions.cs
@@ -17,5 +17,11 @@ namespace DDDEastAnglia.Tests
             var viewName = result.RouteValues["action"];
             return viewName.ToString();
         }
+
+        public static int GetHttpStatusCode(this ActionResult actionResult)
+        {
+            var httpStatusCodeResult = (HttpStatusCodeResult) actionResult;
+            return httpStatusCodeResult.StatusCode;
+        }
     }
 }
--- a/DDDEastAnglia.Tests/ActionResultExtensions.cs
+++ b/DDDEastAnglia.Tests/ActionResultExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.Web.Mvc;
+﻿using System.Net;
+using System.Web.Mvc;
 
 namespace DDDEastAnglia.Tests
 {
@@ -18,10 +19,10 @@ namespace DDDEastAnglia.Tests
             return viewName.ToString();
         }
 
-        public static int GetHttpStatusCode(this ActionResult actionResult)
+        public static HttpStatusCode GetHttpStatusCode(this ActionResult actionResult)
         {
             var httpStatusCodeResult = (HttpStatusCodeResult) actionResult;
-            return httpStatusCodeResult.StatusCode;
+            return (HttpStatusCode) httpStatusCodeResult.StatusCode;
         }
     }
 }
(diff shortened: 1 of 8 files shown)


DDDEastAnglia.Tests/ActionResultExtensions.cs
```diff
@@ -17,5 +17,11 @@ public static string GetRedirectionViewName(this ActionResult actionResult)
             var viewName = result.RouteValues["action"];
             return viewName.ToString();
         }
+
+        public static int GetHttpStatusCode(this ActionResult actionResult)
```
alastairs: Why not return the `HttpStatusCode` from here to save yourself the cast everywhere?

adrianbanks: Yeah, seems sensible.

DDDEastAnglia.Tests/Admin/UserControllerTests.cs
```diff
@@ -0,0 +1,132 @@
...
+using NSubstitute;
+using NUnit.Framework;
+
+namespace DDDEastAnglia.Tests.Admin
+{
+    [TestFixture]
+    public class UserControllerTests
+    {
+        [Test]
+        public void Details_GetsTheCorrectUserDetails()
```
alastairs, line 14: I think you're right that this is testing the implementation: it requires that the method return the value returned from the `IUserProfileRepository`. I think an alternative test might be simply to ensure that the controller calls the appropriate method on `IUserProfileRepository` with the right argument: test the delegation is appropriate rather than the implementation.

adrianbanks, line 14: That is still testing the implementation though. As I had already written the code, I tried to write these test as failing by not looking at the implementation and asserting on the behaviour. Once they were failing, I added the setuip to each test to put the sut in a state where the test could pass..

e.g. I know that calling `controller.Details(123)` should be calling `userProfileRepository.GetUserProfileById(123)`, so ensure that is happening.

As it stands, the actual test (act & assert steps) doesn't really care what happens, so long as it receives a user profile with an id of 123 when it asked for user profile 123. The arrange section of the test is the thing that is implementation specific.

The alternative test would look like:

```
public void Details_GetsTheCorrectUserDetails()
{
    const int userId = 123;
    var userProfileRepository = Substitute.For<IUserProfileRepository>();
    var sessionRepository = Substitute.For<ISessionRepository>();
    var controller = new UserController(userProfileRepository, sessionRepository);

    var actionResult = controller.Details(userId);

    userProfileRepository.Received().GetUserProfileById(userId);
}
```

That doesn't seem any better.

alastairs, line 14: I think your alternative test is a bit better than the original test. J B Rainsberger [talks](http://www.infoq.com/presentations/integration-tests-scam) about two kinds of tests with mocks: collaboration tests and contract tests. IMHO, the alternative test above is a Collaboration test: it checks that the SUT delegates to the mocked service in an appropriate manner. A separate Contract test would, I think, ensure that the mock is configured to behave according to the contract defined by the service interface. In comparison with the original test you wrote, I think this combines both Collaboration and Contract test styles and so is testing more than one thing.

(PS: the video I linked to above is worth watching when you have some time. [This article](http://www.infoq.com/news/2009/04/jbrains-integration-test-scam) sort of summarises some of the key points, with links back to JB's original blog posts, but isn't as good as the video. I would be interested to know what you think about the content of the video, if you get to watch it.)

PR #278, opened 2014-06-17
Add a speakers page in the admin section

adrianbanks:
This is based on #270 as it shares a lot of the styling changes, so that should be merged first.

Add an additional page to show just speakers (i.e. users who have submitted sessions).

### DIFF
--- a/DDDEastAnglia/Areas/Admin/Models/SpeakerModel.cs
+++ b/DDDEastAnglia/Areas/Admin/Models/SpeakerModel.cs
@@ -1,0 +1,23 @@
+﻿using System.ComponentModel;
+
+namespace DDDEastAnglia.Areas.Admin.Models
+{
+    public class SpeakerModel
+    {
+        public int UserId { get; set; }
+
+        [DisplayName("username")]
+        public string UserName { get; set; }
+
+        [DisplayName("name")]
+        public string Name { get; set; }
+
+        [DisplayName("new speaker")]
+        public bool NewSpeaker { get; set; }
+
+        public string GravatarUrl { get; set; }
+
+        [DisplayName("submitted sessions")]
+        public int SubmittedSessionCount { get; set; }
+    }
+}
--- a/DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
+++ b/DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
@@ -7,7 +7,8 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <meta name="description" content="DDD East Anglia conference website 29 June 2013">
         <meta name="author" content="DDD East Anglia">
-        @Styles.Render("~/Content/themes/base/css")
+        @Styles.Render("~/Content/bootstrap")
+        @Styles.Render("~/Content/css")
         @Styles.Render("~/Content/admin")
         @Scripts.Render("~/bundles/modernizr")
         @RenderSection("header", false)
--- a/DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
+++ b/DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
@@ -2,7 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
-        <title>DDD East Anglia Admin @ViewBag.Title</title>
+        <title>DDD East Anglia Admin @(string.IsNullOrWhiteSpace(ViewBag.Title) ? "" : "-") @ViewBag.Title</title>
         <link href="http://www.dddeastanglia.com/favicon.ico" rel="shortcut icon" type="image/x-icon" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <meta name="description" content="DDD East Anglia conference website 29 June 2013">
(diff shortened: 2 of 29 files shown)


DDDEastAnglia/Areas/Admin/Models/SpeakerModel.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.ComponentModel;
+
+namespace DDDEastAnglia.Areas.Admin.Models
+{
+    public class SpeakerModel
+    {
+        public int UserId { get; set; }
+
+        [DisplayName("username")]
```
alastairs, line 9: Shouldn't the display names be specified as they would in the UI, i.e. in Sentence case?

DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
```diff
@@ -2,12 +2,12 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
-        <title>DDD East Anglia Admin @ViewBag.Title</title>
+        <title>DDD East Anglia Admin @(string.IsNullOrWhiteSpace(ViewBag.Title) ? "" : "-") @ViewBag.Title</title>
```
alastairs, line 5: Perhaps this could be pulled out into an HTML helper to avoid duplication of the logic.

DDDEastAnglia/Areas/Admin/Views/User/Delete.cshtml
```diff
@@ -0,0 +1,19 @@
+﻿@model DDDEastAnglia.Models.UserProfile
+
+@{
+    var displayedName = (string.IsNullOrWhiteSpace(Model.Name) ? Model.UserName : Model.Name);
+    ViewBag.Title = string.Format("Delete {0}?", displayedName);
+}
+
+<h2>@ViewBag.Title</h2>
+
+<p>Deleting a user cannot be undone. Are you <em>really</em> sure you want to delete this user?</p>
```
alastairs, line 10: It would be nice if this message were displayed in a modal dialog box rather than a separate page.

DDDEastAnglia/DataAccess/ISpeakerRepository.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.DataAccess
+{
+    public interface ISpeakerRepository
```
alastairs, line 6: This is more of a Query Object than a Repository in many respects. It would be nice to move more of our data access to Query Objects where possible.

DDDEastAnglia/Areas/Admin/Models/SpeakerModel.cs
adrianbanks, line 9: I style these using CSS where appropriate. Personally, I don't like marking fields up like this as it hard-wires display values in the code, but this is how the validation works. For this particular model, we don't (currently) use it to edit, so it is less of an issue.

DDDEastAnglia/Areas/Admin/Views/User/Delete.cshtml
adrianbanks, line 10: Sounds like a nice to have for future work.

DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
adrianbanks, line 5: Possibly, but it is so minimal code I don't think it worth the effort. You would still have the duplication of the call to the helper, which would basically be a very simple wrapper around this method.

DDDEastAnglia/DataAccess/ISpeakerRepository.cs
adrianbanks, line 6: The implementation started out loading data from the database using Simple.Data, and so was a repository. It morphed away from that though as I removed duplication between it and the user profile repository. From the interface point of view, I think it should be consistent with the others (i.e. either they are are  query objects or all repositories).

I might rename all of the repository interfaces to something more query-based (although I don't like that term as it has quite database-specific connotations).

DDDEastAnglia/Areas/Admin/Models/SpeakerModel.cs
alastairs, line 9: Fair enough, sounds like a good approach.

DDDEastAnglia/Areas/Admin/Views/Shared/_Layout.cshtml
alastairs, line 5: Oh, no, I meant the whole title string. I hadn't noticed it was in the `_Layout.cshtml` file, though, so I thought that this was going to be duplicated. Obviously that won't actually be the case, so let's forget I said anything :wink:

DDDEastAnglia/DataAccess/ISpeakerRepository.cs
alastairs, line 6: > The implementation started out loading data from the database using Simple.Data, and so was a repository

That doesn't necessarily follow. See this example loading data from RavenDB: http://lostechies.com/jimmybogard/2012/10/08/favor-query-objects-over-repositories/

> From the interface point of view, I think it should be consistent with the others (i.e. either they are are query objects or all repositories).

That is a fair point. Perhaps leave it for now and we can take a look at refactoring the repositories to query objects another time.

alastairs:
This is good to merge. :+1:

PR #281, opened 2014-06-18
Disable parts of the timeline as the dates for them pass

adrianbanks:
As each date in the timeline passes, it will show as "disabled" to show that that item has passed. I've styled it as such:

### DIFF
--- a/DDDEastAnglia.Tests/Controllers/TimelineControllerTests.cs
+++ b/DDDEastAnglia.Tests/Controllers/TimelineControllerTests.cs
@@ -3,6 +3,7 @@ using DDDEastAnglia.Controllers;
 using DDDEastAnglia.DataAccess;
 using DDDEastAnglia.DataAccess.SimpleData.Models;
 using DDDEastAnglia.Domain.Calendar;
+using DDDEastAnglia.Helpers;
 using DDDEastAnglia.Models;
 using NSubstitute;
 using NUnit.Framework;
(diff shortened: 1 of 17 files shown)


DDDEastAnglia/LocalDateTimeOffsetProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace DDDEastAnglia
+{
+    public class LocalDateTimeOffsetProvider : IDateTimeOffsetProvider
+    {
+        public DateTimeOffset CurrentDateTime()
+        {
+            return DateTimeOffset.Now;
```
alastairs, line 9: Should this be `UtcNow` instead for time zone compatibility? Perhaps that's brought by the `DateTimeOffset` instead, though.

DDDEastAnglia/Views/Timeline/_Timeline.cshtml
```diff
@@ -3,22 +3,22 @@
 <section>
     <h1>Timeline</h1>
     <dl>
-        <dt><strong>@Model.SubmissionOpens:</strong></dt>
-        <dd>Session submission opens</dd>
+        <dt class="@(Model.SubmissionPeriodPassed ? "passedTimelinePeriod" : "")"><strong>@Model.SubmissionOpens:</strong></dt>
```
alastairs: There's a lot of duplicated logic here, so it would be nice to have this whole timeline item in a helper, e.g. `@Html.TimelineItem(periodPassed, periodStartDate, label)`.

DDDEastAnglia/LocalDateTimeOffsetProvider.cs
adrianbanks, line 9: Well for this class, no, as that would mean that it is a `UtcDateTimeOffsetProvider`. :wink: 

Getting `DateTimeOffset.Now` gives a local time, but with the correct offset, meaning that TimeZone issues don't come into play, but making it easier to work with as you are always in the correct time/time zone.

DDDEastAnglia/Views/Timeline/_Timeline.cshtml
adrianbanks: I debated doing that, but then there would still be lots of duplicated calls hiding a tiny amount of logic. I didn't think it worth adding the complexity of yet another extension method just for this one page. I can give it a try though.

alastairs: I don't think it is a tiny amount of logic: the helper would wrap both the `dt` and the `dd`, so it also wraps all the formatting. (This is logic of a kind too.) I'd like to see it attempted at least.

adrianbanks: I've split the logic into a partial view. There is a lot more to it compared with before, but it gave me the chance to refactor some of the logic out of the controller.

I'm not keen on the use of `ViewDataDictionary` in `_Timeline.cshtml`, but I can't think of another way to keep the presentation strings in the view.

DDDEastAnglia.Tests/Controllers/TimelineControllerTests.cs
```diff
@@ -0,0 +1,186 @@
...
+namespace DDDEastAnglia.Tests.Controllers
+{
+    [TestFixture]
+    public sealed class TimelineControllerTests
+    {
+        [Test]
+        public void Details_ShowsThatTheSubmissionPeriodHasNotPassed_WhenTheCurrentDateIsBeforeTheSubmissionPeriod()
+        {
+            // 30 mins before session submission opens
+            var controller = SetUpController(CreateDateTime(2010, 3, 31, 23, 30, 0));
```
alastairs: It would be nicer, perhaps, to use a builder pattern for this stuff? E.g.: 

``` csharp
AController().AtTime(30.MinutesBefore(SessionSubmissionOpens)).Build();
```

PR #284, opened 2014-06-22
Don't differentiate between valid and invalid user details when resetting passwords

adrianbanks:
Currently, you can brute force the site to get a valid username and/or email address as we return the message "Could not reset password" if the user doesn't exist, as opposed to a page that tells you to check your email. This changes that behaviour to not divulge whether or not we have found the user's details.

This is not ideal, as it means that if you guess at your username/password and get it wrong, you will get no reset password email. The correct way to deal with it is to lock the user (ip address?) out after several failed tries, but that it a much bigger and more complex change.


alastairs:
Looks like you've deleted tests here, and so the metric is (obviously) failing. This is fine, and should be fixed once it's merged.

PR #287, opened 2014-06-27
Fix #285 Voting isn't working on the staging site

### LINKED ISSUE #285: Voting isn't working on the staging site
When you click on the vote button, an ajax request is sent, but nothing appears to happen.

It looks like each time I click the button, I'm getting a new cookie so the site thinks that I've not already voted.

adrianbanks:
The cookie used to store the votes had an expiry date in the past. I've made it pull the expiry date from the database based on the real conference dates.


alastairs:
Good find. Now you describe what it was, it makes perfect sense for the behaviour you described! I've looked over it a bit, but not in full yet. I'll try and get onto it later this afternoon.

adrianbanks:
It was Ian who gave me a good steer to find it.

philpursglove:
Teamwork FTW!

----- Original Message -----

alastairs:
@philpursglove :+1: :smile:

adrianbanks:
> Before the refactoring, it made sense within the domain - the method to see if a vote had been logged for a particular session in that cookie.... I feel as though information and behaviour has been lost

Before, there was a class that did something that looked sensible, but wasn't actually used anywhere. The only bits that were used were the two properties (which are still on the refactored class).

The refactored class needs more logic to get these two values (name and expiry), as they are no longer hard-coded. To get them requires database access, hence the dependencies. I did consider making them method calls as they had logic in them, but then that looked weird in usage so left it as is.

alastairs:
> The refactored class needs more logic to get these two values (name and expiry), as they are no longer hard-coded. To get them requires database access, hence the dependencies.

I would prefer that `VotingCookie` became a true Value Object (ideally immutable too), and that something else was responsible for the necessary logic to create them. The interface can then be deleted, because `VotingCookie`s will be easy to construct again, with the two pieces of information they expose (name and expiry date).

adrianbanks:
I've been round in circles with this a few times now due to needing some information from the cookie before I can load the cookie. I think I've now sorted it.

PR #297, opened 2014-07-04
Show a user's login methods on the user details page in the admin area

adrianbanks:
We have had a few people who can't login because they originally used a provider (e.g. GitHub), but are then trying to log in via another mechanism. This change shows a user's available login methods in the admin section, so we can easily point them in the right direction.

### DIFF
--- a/DDDEastAnglia.Tests/Admin/UserControllerTests.cs
+++ b/DDDEastAnglia.Tests/Admin/UserControllerTests.cs
@@ -10,13 +10,13 @@ namespace DDDEastAnglia.Tests.Admin
     [TestFixture]
     public class UserControllerTests
     {
+        private IUserProfileRepository userProfileRepository;
+
         [Test]
         public void Details_GetsTheCorrectUserDetails()
         {
             const int userId = 123;
-            var userProfileRepository = Substitute.For<IUserProfileRepository>();
-            var sessionRepository = Substitute.For<ISessionRepository>();
-            var controller = new UserController(userProfileRepository, sessionRepository);
+            var controller = CreateController();
 
             controller.Details(userId);
 
@@ -26,9 +26,7 @@ namespace DDDEastAnglia.Tests.Admin
         [Test]
         public void Details_ReturnsA404_WhenTheUserCannotBeFound()
         {
-            var userProfileRepository = Substitute.For<IUserProfileRepository>();
-            var sessionRepository = Substitute.For<ISessionRepository>();
-            var controller = new UserController(userProfileRepository, sessionRepository);
+            var controller = CreateController();
 
             var actionResult = controller.Details(123);
 
@@ -39,9 +37,7 @@ namespace DDDEastAnglia.Tests.Admin
         public void Edit_GetsTheCorrectUserDetails()
         {
             const int userId = 123;
-            var userProfileRepository = Substitute.For<IUserProfileRepository>();
-            var sessionRepository = Substitute.For<ISessionRepository>();
-            var controller = new UserController(userProfileRepository, sessionRepository);
+            var controller = CreateController();
 
             controller.Edit(userId);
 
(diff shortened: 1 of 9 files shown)


DDDEastAnglia/DataAccess/IAccountLoginMethodQuery.cs
```diff
@@ -0,0 +1,9 @@
+﻿using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.DataAccess
+{
+    public interface IAccountLoginMethodQuery
```
alastairs, line 5: Shame that this interface has only one implementation. It might be better as an `IQuery<T, TReturn>`: this is the route I'm navigating with my query objects refactoring branch, at least.

DDDEastAnglia/DataAccess/SimpleData/Queries/AccountLoginMethodQuery.cs
```diff
@@ -0,0 +1,31 @@
...
+            var dddeaLogin = db.webpages_Membership.FindByUserId(userId);
+            bool hasDddeaLogin = dddeaLogin != null;
+            
+            List<string> oauthLogins = db.webpages_OAuthMembership.FindAllByUserId(userId)
+                                         .Select(db.webpages_OAuthMembership.Provider).ToScalarList<string>();
+            bool hasGitHubLogin = oauthLogins.Contains("github");
+            bool hasTwitterLogin = oauthLogins.Contains("twitter");
+            bool hasGoogleLogin = oauthLogins.Contains("google");
+            
+            return new LoginMethods
```
alastairs, line 22: I'm not 100% convinced by this (it's just a bag of `bool`s and doesn't really represent a concept), but we can probably leave it for now. 

A nicer design might be to return an `IEnumerable<LoginMethod>`, perhaps.

DDDEastAnglia/DataAccess/IAccountLoginMethodQuery.cs
adrianbanks, line 5: Possibly, although you can't bind that in Ninject using convention (unless I've missed something). The Ninject bindings need looking at anyway though, as there is some oddity with them.

DDDEastAnglia/DataAccess/SimpleData/Queries/AccountLoginMethodQuery.cs
adrianbanks, line 22: Possibly, although I think you may then need to put display properties into the `LoginMethod` class (i.e. css class for the icon, tooltips, order, etc...) to render it.

adrianbanks, line 22: I've had a go at doing this. Ultimately, somewhere you need to map between the provider name and the display properties (e.g. name, css icon). See what you think: https://github.com/adrianbanks/dddeastanglia/commit/9b8a4717085281a6b73149df1d235e37b68af1e7

adrianbanks, line 22: @alastairs Based on your comments on that last diff, I've taken it further: [diff](https://github.com/dddeastanglia/DDDEastAnglia/compare/9b8a4717085281a6b73149df1d235e37b68af1e7...5b482efdaec7101387ce9b7e735a8cd5e5f3b789). See what you think.

DDDEastAnglia.Tests/Admin/UserControllerTests.cs
```diff
@@ -10,13 +10,13 @@ namespace DDDEastAnglia.Tests.Admin
     [TestFixture]
     public class UserControllerTests
     {
+        private IUserProfileRepository userProfileRepository;
```
alastairs, line 13: Why is this a field? It doesn't appear to need to be.

adrianbanks, line 13: Because several of the tests use it to assert behaviour. Those lines don't show up in this diff as I haven't changed them.

DDDEastAnglia/DataAccess/SimpleData/Queries/AccountLoginMethodQuery.cs
alastairs, line 22: The updated version looks good :smile:

adrianbanks, line 22: I'll close this and open a new one from the other branch.

PR #299, opened 2014-07-05
Fix #289 Use the correct number of sessions for the conference

### LINKED ISSUE #289: Store extra information about the conference
A lot of the pages in the voting area use the number of total sessions when displaying the graphs. It would be good to get this from the database, from information associated with the conference.

adrianbanks:
This change adds the total number of time slots and the total number of tracks to the database, and uses those values in the voting pages to visualise the data with the correct number of total sessions.

After this is merged, the number of time slots and tracks will both be set to 5 in the database.


DDDEastAnglia.DatabaseMigrations/Migrations/2-AddTrackInformationToConferencesTable.cs
```diff
@@ -0,0 +1,24 @@
...
+    [Migration(20140704)]
+    public class AddTrackInformationToConferencesTable : Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("Conferences")
+                    .AddColumn("NumberOfTimeSlots").AsInt32().NotNullable().WithDefaultValue(0)
+                    .AddColumn("NumberOfTracks").AsInt32().NotNullable().WithDefaultValue(0);
+
+            Execute.Sql("UPDATE [dbo].[Conferences] SET [NumberOfTimeSlots] = 5, [NumberOfTracks] = 5");
```
alastairs, line 14: There's no `Update.Table()` or equivalent? Perhaps that makes sense as it's not DDL.

adrianbanks, line 14: No. Only schema changes are supported in the fluent style. I don't technically require this line, but everyone would have to manually edit their databases by hand to insert this, so I included it for ease.

alastairs, line 14: Fair enough. Maybe set the "Number of Time Slots" default value to 5? We don't vary this as much as the number of tracks. I think that makes sense in our domain, too. What do you think?

adrianbanks, line 14: I had considered that, but then it's a bit weird as to why the schema says it is 5 with no way to document the reason. It might be better to have the default set when a new conference is created, then it can be documented in code.

alastairs, line 14: Okey-dokey, makes sense.

adrianbanks:
> AutoFixture would be (y)our friend here more than builders, IMO.

Not sure how you mean. The `VotingControllerBuilder` is used to create the SUT where it is used, and so can't be made with AutoFixture.

The other builders are used to provide canned answers, so I need to know what those answers are - unless there is a way with AutoFixture to do that (but wouldn't that defeat the purpose of using it in the first place)?

alastairs:
So AutoFixture can act as a SUT Factory and provide instances (optionally mocked) of the SUT's dependencies, so we wouldn't need VotingControllerBuilder. As far as the canned answers are concerned... perhaps you would still need builders for those. I have mock support on the list for [Bob](https://github.com/alastairs/BobTheBuilder); fancy spiking it? :wink:

alastairs:
:+1: Needs manual merge

PR #301, opened 2014-07-08
Update manual migration script

adrianbanks:
This upgrades the 2013 database to the correct 2014 state (including the fix I did earlier tonight).


alastairs:
Why does this have all the search stuff in it too?

adrianbanks:
It shouldn't - I'll rebase it later.

adrianbanks:
Sorted.

PR #307, opened 2014-07-17
Remove lots of unused libraries

adrianbanks:


alastairs:
The compilation error on this branch is:

```
[04:52:52][Step 1/1]        (MvcBuildViews target) -> 
[04:52:52][Step 1/1]          error ASPRUNTIME : The pre-application start initialization method Start on type System.Web.WebPages.Administration.PreApplicationStartCode threw an exception with the following error message: Could not load file or assembly 'NuGet.Core, Version=1.6.30117.9648, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.. [C:\TeamCity\buildAgent\work\77c9f48fc195cd03\DDDEastAnglia\DDDEastAnglia.csproj]
```

alastairs:
Why did you remove Glimpse? Is it because it's configured incorrectly?

adrianbanks:
Because we don't really use it and it was adding a payload to the pages. Also, turning it in didn't seem to actually do anything, so I suspect it wasn't configured correctly.

adrianbanks:
> The compilation error on this branch is...

Do you happen to know what we were using Nuget.Core.dll for? It wasn't used anywhere, so I'm surprised that the compilation fails as a result.

alastairs:
> Do you happen to know what we were using Nuget.Core.dll for?

_We_ don't use it, but I suspect MVC does. The same goes for EF, actually: when you create a new MVC project, you get EF with it because of the Membership Provider stuff (I think - I can't remember exactly).

It's the `MvcBuildViews` target that fails, so I suspect MVC does the NuGet Package Restore magic during this process as well as earlier compilation steps.

adrianbanks:
It is safe to remove EF - it is just there in the default project template for some reason. 

NuGet.Core isn't part of any standard project template, so something must have added it explicitly. It is for messing with Nuget references programmatically, so it's a bit odd that we have it as a reference in the project.

alastairs:
According to the NuGet Package Visualizer:

temp

A quick Find References on the `WebPages.Administration` assembly suggests that this NuGet package is also unused and could be removed.

adrianbanks:
I'll look at that tonight. It did all build locally on my machine, but that may be MSBuild doing its "find a reference from somewhere" routine.

alastairs:
Nope, I think MvcBuildViews might only run in Release configuration, or something like that.

adrianbanks:
Ok, this now builds on Team City.

PR #308, opened 2014-07-17
Add some new pages to the voting stats to enable us to see who has voted for each session

adrianbanks:


DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
```diff
@@ -210,5 +210,34 @@ public ActionResult DuplicateVotes()
...
+
+        public ActionResult VotersForSessions()
+        {
+            var leaderboardSessions = dataProvider.GetLeaderBoard(int.MaxValue, true);
+            return View(leaderboardSessions);
+        }
+
+        [HttpPost]
+        [AllowCrossSiteJson]
+        public ActionResult GetUsersWhoVotedForSession(int sessionId)
```
alastairs, line 218: Why do the POST requests start with "Get"?

DDDEastAnglia/Areas/Admin/Views/Voting/IPAddresses.cshtml
```diff
@@ -24,7 +24,7 @@
             @Html.DisplayFor(modelItem => item.IPAddress)
             
             <div>
-                <a id="link@(item.IPAddress.Replace(".", "-"))" href="#" onclick="javascript:resolveIP('link@(item.IPAddress.Replace(".", "-"))', '@item.IPAddress', 'resolved@(item.IPAddress.Replace(".", "-"))');return false;" title="Resolve IP address"><i class="icon-search"></i></a>
+                <a id="iplink@(item.IPAddress.Replace(".", "-"))" href="#" onclick="javascript:resolveIP('iplink@(item.IPAddress.Replace(".", "-"))', '@item.IPAddress', 'resolved@(item.IPAddress.Replace(".", "-"))');return false;" title="Resolve IP address"><i class="icon-search"></i></a>
```
alastairs: `item.IPAddress.Replace(".", "-")` is repeated three times in this line. An extension method would be useful to encapsulate this logic.

DDDEastAnglia/Areas/Admin/Views/Voting/IpAddressesThatVotedForSessions.cshtml
```diff
@@ -0,0 +1,43 @@
+﻿@model IList<DDDEastAnglia.VotingData.Models.SessionLeaderBoardEntry>
+
+<h2>Votes from IP addresses for each session</h2>
+
+<p>This shows which IP addresses have voted for each session.</p>
+
+<table class="table table-striped table-hover table-condensed">
+    <tbody>
```
alastairs, line 8: A `<thead />` would be useful to give titles to the columns.

DDDEastAnglia/Areas/Admin/Views/Voting/VotersForSessions.cshtml
```diff
@@ -0,0 +1,37 @@
+﻿@model IList<DDDEastAnglia.VotingData.Models.SessionLeaderBoardEntry>
+
+<h2>Voters for each session</h2>
+
+<p>This shows who voted for each session.</p>
+
+<table class="table table-striped table-hover table-condensed">
+    <tbody>
```
alastairs, line 8: Again, a <thead /> would be useful here.

DDDEastAnglia/Areas/Admin/Views/Voting/IpAddressesThatVotedForSessions.cshtml
adrianbanks, line 8: I intentionally didn't put titles on the columns because the table contains nested tables, so the column titles don't really make sense when the nested tables are expanded.

DDDEastAnglia/Areas/Admin/Views/Voting/VotersForSessions.cshtml
adrianbanks, line 8: Same reason as before.

DDDEastAnglia/Areas/Admin/Controllers/VotingController.cs
adrianbanks, line 218: This _is_ getting users, but it is a POST because it is an ajax call done via javascript.

alastairs, line 218: AJAX calls are just regular HTTP requests, though, so you could quite easily make it a GET rather than a POST. This would be more semantically correct and RESTful.

adrianbanks, line 218: It didn't work when I tried switching it to a get request before. I can't remember what I did though so I'll have another try.

PR #310, opened 2014-07-21
Show stats after voting closes

adrianbanks:
Otherwise, we can't see the voting results once voting closes. This change keeps the voting stats visible until the agenda publish date has passed.


adrianbanks:
Merging this as it **needs** to be in production before the end of Friday, otherwise we can't easily see the results of voting.

PR #317, opened 2014-08-11
Update the accommodation page

adrianbanks:
This pull request removes the warning about graduation, and adds some more hotels to the list, including a map showing the location of them all.


adrianbanks:
Merging this as it is a bit pointless once all of the tickets are gone.

PR #319, opened 2014-08-22
Added drinks reception to agenda

philpursglove:
Added a line to the Agenda for the Red Gate reception

### DIFF
--- a/DDDEastAnglia/Views/Home/Agenda.cshtml
+++ b/DDDEastAnglia/Views/Home/Agenda.cshtml
@@ -247,7 +247,7 @@
     
     <tr class="nonSession">
         <td class="timeCell">17:00</td>
-        <td colspan="5">Drinks Reception (sponsored by <a href="http://www.red-gate.com/">Red Gate)</a></td>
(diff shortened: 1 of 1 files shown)


DDDEastAnglia/Views/Home/Agenda.cshtml
```diff
@@ -247,6 +247,11 @@
     
     <tr class="nonSession">
         <td class="timeCell">17:00</td>
+        <td colspan="5">Drinks Reception (sponsored by <a href="http://www.red-gate.com/">Red Gate)</a></td>
```
adrianbanks: Red Gate gave us a special url to link back to their site (http://redg.at/11xaIPC). We should probably use that here too.

alastairs: That goes to the home page. I've asked the People Team whether they want us to link directly into the careers section of the site, awaiting confirmation. 

(Also that link is for last year's conference, so I'll let the Agency know we're still using that.)

alastairs:
Having spoken to Jodie, the URL she would like us to use is http://dev.red-gate.com/.

PR #324, opened 2014-12-14
Refactor of the external logins 

adrianbanks:
Had this branch lying around for a while. The changes decouple the OAuth web security stuff from our code, and also allow the icons for the OAuth providers to be configured in the app settings.


DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace DDDEastAnglia.Tests.Helpers.LoginMethods
+{
+    [TestFixture]
+    public sealed class LoginMethodIconProviderTests
+    {
+        [Test]
+        public void GetIcon_ReturnsTheConfiguredIconForOAuthProvider()
+        {
+            var appSettingsProvider = Substitute.For<IAppSettingsProvider>();
+            appSettingsProvider.GetSetting(LoginMethodIconProvider.AppSettingsKeyPrefix + "test").Returns("test-icon");
```
alastairs: Do all consumers of the `IAppSettingsProvider` have to know about (and remember to concatenate) the `AppSettingsKeyPrefix` from the `LoginMethodIconProvider`? It seems a bit like Countdown Code - "two from the top and three of your choice please Carol". It also feels a bit like you're replicating the behaviour of the real `IAppSettingsProvider` in this Substitute. 

Perhaps this is unavoidable, though - is `IAppSettingsProvider` one of our types, or a ASP.NET/BCL type?

DDDEastAnglia/Controllers/AccountController.cs
```diff
@@ -16,18 +17,41 @@ namespace DDDEastAnglia.Controllers
     public class AccountController : Controller
     {
         private readonly IUserProfileRepository userProfileRepository;
-
-        public AccountController(IUserProfileRepository userProfileRepository)
+        private readonly IBuild<LoginMethod, LoginMethodViewModel> loginMethodViewModelBuilder;
+        private readonly ExternalLoginsProvider externalLoginsProvider;
+        private readonly IOAuthClientInfo oAuthClientInfo;
+
+        public AccountController(IUserProfileRepository userProfileRepository,
```
alastairs, line 24: This has a lot of dependencies. It looks to me (from this constructor signature alone, I've not yet delved in in any more detail) as though the account controller needs two things:
1. Something to convert a `LoginMethod` to a `LoginMethodViewModel`.
2. Something to complete an authentication request, returning a representation of a User.

Ok, so maybe it's three things and there are two controllers to be split out: one that handles authentication, one that handles Users (e.g. profile data).

DDDEastAnglia/Controllers/AccountController.cs
```diff
@@ -141,70 +161,78 @@ public ActionResult ChangePassword(LocalPasswordModel model)
...
         }
 
         [ChildActionOnly]
-        public ActionResult RemoveExternalLogins()
+        public ActionResult ExternalLogins(string returnUrl)
         {
-            var accounts = OAuthWebSecurity.GetAccountsFromUserName(User.Identity.Name);
-            var externalLogins = new List<ExternalLogin>();
+            var allExternalLogins = externalLoginsProvider.GetAllAvailable();
+            var usersExternalLogins = externalLoginsProvider.GetForUser(User.Identity.Name).ToList();
```
alastairs: I don't think I'd bother with this intermediary list; the dictionary is sufficient and it just adds more noise.

DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
adrianbanks: > Do all consumers of the IAppSettingsProvider have to know about (and remember to concatenate) the AppSettingsKeyPrefix from the LoginMethodIconProvider?

No, only these tests. All of the icon settings are prefixed to group them in the appsettings. It's only these tests that need to add the prefix to the setting (as the login icon code is expecting to get a setting with that prefix).

DDDEastAnglia/Controllers/AccountController.cs
adrianbanks, line 24: It's more involved than that.

The actions for registering a user and changing a password only require one dependency each (`Register` -> `IUserProfileRepository`, `ChangePassword` -> `IOAuthClientInfo`), but the external login management actions require all four of the dependencies (`IUserProfileRepository`, `IOAuthClientInfo`, `ExternalLoginsProvider` and `IBuild<LoginMethod, LoginMethodViewModel>`). 

The external login actions need the dependencies because they pull together several concepts related to logins, users and security.

DDDEastAnglia/Controllers/AccountController.cs
adrianbanks: It was useful for debugging when it wasn't working correctly. It is also used later in the method (although that usage could be replaced with a call to the dictionary as well, but I don't think it is as clear).

DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
alastairs: Ok. Could we factor that logic out of the tests then, so they're clearer to read? Like an `AppSettingsProviderBuilder`, perhaps (or just a Factory Method). Actually, it seems to me as though the AppSettingsProvider could just return `"test-icon"` for all input and the test would still pass - that's fine (we're testing that the SUT uses the return value from the Substitute correctly), and embracing it allows us to remove cruft from the test.

alastairs: The flip side of that last remark is that we need to test the `AppSettingsProvider` is _called_ correctly too - the appropriate method is called, with the expected arguments.

DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
```diff
@@ -0,0 +1,35 @@
...
+            var iconProvider = new LoginMethodIconProvider(appSettingsProvider);
+
+            string icon = iconProvider.GetIcon("test");
+
+            Assert.That(icon, Is.EqualTo("test-icon"));
+        }
+
+        [Test]
+        public void GetIcon_ReturnsTheUnknownIconForAnUnknownOAuthProvider()
```
alastairs, line 23: The same comments seem to apply to this test AFAICT.

DDDEastAnglia/Controllers/AccountController.cs
alastairs, line 24: Ok. Is there a service we can pull out to abstract over some of these dependencies for the external login actions? I can see the controller needing this façade service and the `IBuild<LoginMethod, LoginMethodViewModel>`, but not much else. 

Ugh, actually, this looks like a fat controller that needs breaking up, but this isn't the place for it.

DDDEastAnglia/Controllers/AccountController.cs
```diff
@@ -141,70 +161,78 @@ public ActionResult ChangePassword(LocalPasswordModel model)
...
+                bool hasLocalAccount = oAuthClientInfo.HasLocalAccount(WebSecurity.GetUserId(User.Identity.Name));
+                var externalLogins = oAuthClientInfo.GetAccountsFromUserName(User.Identity.Name);
 
-                    if (hasLocalAccount || OAuthWebSecurity.GetAccountsFromUserName(User.Identity.Name).Count > 1)
-                    {
-                        OAuthWebSecurity.DeleteAccount(provider, providerUserId);
-                        scope.Complete();
-                        message = "The external login has been removed successfully.";
-                    }
+                if (hasLocalAccount || externalLogins.Count > 1)
```
alastairs, line 226: This logic appears to be duplicated - or at least there's something _very_ similar higher up. Can we extract a method to put a name on it?

DDDEastAnglia/Helpers/LoginMethods/ExternalLoginsProvider.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using DDDEastAnglia.Helpers.Security;
+using DDDEastAnglia.Models;
+using Microsoft.Web.WebPages.OAuth;
+
+namespace DDDEastAnglia.Helpers.LoginMethods
+{
+    public class ExternalLoginsProvider
```
alastairs: Can we put a better name on this - maybe it's an `ExternalLoginsDirectory`?

DDDEastAnglia/Helpers/LoginMethods/ExternalLoginsProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            return externalLogins.OrderBy(l => l.DisplayName);
+        }
+
+        public IEnumerable<LoginMethod> GetAllAvailable()
+        {
+            var authenticationClientDatas = oAuthClientInfo.RegisteredClientData();
+            var externalLogins = new List<LoginMethod>();
+
+            foreach (var externalLogin in authenticationClientDatas)
```
alastairs: Seems like there's some duplicated code between these methods. Is it possible to pull out a helper method to reduce that? [Not necessarily, I know; the code does look just enough different, but perhaps it can be munged to make it work.]

DDDEastAnglia/Helpers/LoginMethods/ExternalLoginsProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+                var provider = externalLogin.AuthenticationClient.ProviderName;
+
+                externalLogins.Add(new LoginMethod
+                {
+                    DisplayName = externalLogin.DisplayName,
+                    ProviderName = provider
+                });
+            }
+
+            return externalLogins.OrderBy(l => l.DisplayName);
```
alastairs: Is it worth doing the ordering up-front on the `foreach`, and then `yield return`ing each element from the method, rather than building a list and returning that? Given the size of the list, perhaps not, but worth a thought at least.

DDDEastAnglia/Helpers/LoginMethods/LoginMethodIconProvider.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using DDDEastAnglia.Helpers.AppSettings;
+
+namespace DDDEastAnglia.Helpers.LoginMethods
+{
+    public class LoginMethodIconProvider
```
alastairs, line 6: Could this become a `LoginMethodIcon`? Or perhaps it should just be a property of a `LoginMethod` in general.

DDDEastAnglia/Helpers/Security/IOAuthClientInfo.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Collections.Generic;
+using DotNetOpenAuth.AspNet;
+using Microsoft.Web.WebPages.OAuth;
+
+namespace DDDEastAnglia.Helpers.Security
+{
+    public interface IOAuthClientInfo
```
alastairs, line 7: Wow, ok, that's a big interface. Given that it's defining a wrapper around a static class, it should be über-easy to turn this into a set of [Role Interfaces](http://martinfowler.com/bliki/RoleInterface.html) to better adhere to the Interface Segregation Principle.

DDDEastAnglia/Views/Account/_ExternalLoginMethodsPartial.cshtml
```diff
@@ -0,0 +1,12 @@
+﻿@model IEnumerable<DDDEastAnglia.Models.LoginMethodViewModel>
+
+@foreach (var externalLogin in Model)
+{
+    <li>
+        @using (Html.BeginForm("ExternalLogin", "Account", new { provider = externalLogin.ProviderName, returnUrl = Url.Action("Index", "Home") }, FormMethod.Post, new { id = externalLogin.ProviderName + "Loginform" }))
```
alastairs, line 6: Why is the `returnUrl` hard-coded here, rather than being set from the `ViewBag`?

DDDEastAnglia/Views/Account/_ExternalLoginMethodsPartial.cshtml
```diff
@@ -0,0 +1,12 @@
...
+        {
+            @Html.AntiForgeryToken()
+            <a href="javascript:document.getElementById('@externalLogin.ProviderName' + 'Loginform').submit();"><i class="@externalLogin.Icon"></i> Log in with @externalLogin.Name</a>
```
alastairs, line 9: Should these be buttons rather than links? I can't work out where this is used, and there aren't any clues in the filename or path. Maybe it's in the menu?

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
```diff
@@ -0,0 +1,49 @@
+﻿@model IEnumerable<DDDEastAnglia.Models.LoginMethodViewModel>
```
alastairs, line 1: This is really baking my noodle. We've had `ExternalLoginButtonsPartial`, `ExternalLoginMethodsPartial`, `ExternalLoginsListPartial` (deleted) and now `ExternalLoginsPartial`. All of them under /Views/Account. What do these things do, where do they appear, etc., etc.?

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
```diff
@@ -0,0 +1,49 @@
...
+
+@if (Model.Any())
+{
+    foreach (var loginMethod in Model)
+    {
+        <div style="margin-bottom: 10px;">
```
alastairs, line 7: Can we kill these in-line styles in favour of CSS classes?

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
```diff
@@ -0,0 +1,49 @@
...
+                @if (loginMethod.ProviderUserId == null)
+                {
+                    using (Html.BeginForm("ExternalLogin", "Account", new {ViewBag.ReturnUrl}))
+                    {
+                        @Html.AntiForgeryToken()
+                        <button type="submit" 
+                                class="btn btn-success" 
+                                name="provider"
+                                value="@loginMethod.ProviderName"
+                                title="Log in using @loginMethod.Name">
```
alastairs, line 22: This should read something like "Add @loginMethod.Name to your account"

DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
alastairs: Furthermore, it wasn't until I actually saw it in the config file that I realised that this didn't feel right to me: appSettings is intended for configuration data; the icon associated with a login method isn't configuration data, it's just a user display thing. Enabling/disabling a login method - ok for appSettings; defining how a login method should be displayed to the user - not ok for appSettings.

alastairs:
Finally got to the end of this review - sorry it took so long! Overall it looks like a good change that will improve this part of the site.

DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
adrianbanks: I tend to agree (although appSettings are really just pairs of keys and values), but in this case I think it buys us something. The [original version of the icon provider](https://github.com/adrianbanks/dddeastanglia/blob/de683c0e21dfbfaf8f30210a7b27370a72d6e9af/DDDEastAnglia/Helpers/LoginMethods/LoginMethodIconProvider.cs) has bootstrap icon names embedded in the C#, which didn't feel right. Doing it this way means that we can change them without the need to recompile.

DDDEastAnglia/Controllers/AccountController.cs
adrianbanks, line 24: Agree - that's why I didn't do it here.

DDDEastAnglia/Helpers/LoginMethods/ExternalLoginsProvider.cs
adrianbanks: I don't think there is a way to do it without making the code much harder to read.

DDDEastAnglia/Helpers/Security/IOAuthClientInfo.cs
adrianbanks, line 7: I just extracted the interface (I think) to help me with testing. The OAuth code is very messy, and could do with looking at separately.

DDDEastAnglia/Views/Account/_ExternalLoginMethodsPartial.cshtml
adrianbanks, line 6: Not sure I follow what you are getting at.

DDDEastAnglia/Views/Account/_ExternalLoginMethodsPartial.cshtml
adrianbanks, line 9: No, they are used in the menu so need to be links.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
adrianbanks, line 1: I agree it is a bit confusing. This is what they are:

`_ExternalLoginsPartial`: the login methods that are shown on the manage logins page
`_ExternalLoginMethodsPartial`: the login methods that are shown on the login menu
`_ExternalLoginButtonsPartial`: the login methods that are shown on the login page

They are all slightly different in what they do and how they display.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
adrianbanks, line 7: They could, but I don't see a huge value in doing so as they are only to style one small area of one page.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
adrianbanks, line 22: That doesn't really make sense. By clicking the button, you are allowing yourself to log in with the provider you have clicked on (e.g. by clicking on the GitHub button, I am allowing myself to log into the DDDEA site using GitHub).

alastairs, line 22: Oh, I thought this partial was used on the profile page to allow you to add more external logins to your account. If that's not the case, and it's used from the login menu or similar, then I agree the current wording makes more sense.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
alastairs, line 7: It's a coding standards/code hygiene thing. It's much harder to keep inline styles clean and maintainable than CSS classes. The added benefit of putting it in a class is that it allows us to give the style a meaningful (i.e. semantic) name.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
alastairs, line 1: Why are the last two separate? Could they not be the same, just rendered slightly differently? Perhaps that's beyond the capabilities of our current CSS (i.e., perhaps it requires us to pull in Sass/LESS to achieve that kind of extensibility). Perhaps we could achieve the same with a little bit of logic in the view to determine where it's being shown - horrid, I know, but would remove the duplication pending a refactor to a better solution. 

How about something like:

`_ManageLoginMethods` - or `_ManageExternalLoginMethods` if adding "external" provides much value.
`_LoginMethodsList`

DDDEastAnglia/Views/Account/_ExternalLoginMethodsPartial.cshtml
alastairs, line 6: Elsewhere the `returnUrl` is set via the `ViewBag`, but here it is hard-coded to the homepage. That doesn't seem right to me - someone might be on the submissions page, choose to log in to submit, and end up on the home page again.

DDDEastAnglia/Helpers/Security/IOAuthClientInfo.cs
alastairs, line 7: Ok. I still think it would benefit from being split into a number of interfaces though.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
adrianbanks, line 1: They are both different views of the same model, used for different purposes. They have a differing HTML structure, so it would be messy to combine the two.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
adrianbanks, line 7: Any suggestions for a name? I couldn't come up with anything that was meaningful, which is part of the reason why I left them inline.

DDDEastAnglia/Controllers/AccountController.cs
alastairs, line 24: Ok, well a façade service or two will help with later refactoring, and shouldn't be too hefty to implement, so can we try to get that in at least?

DDDEastAnglia/Helpers/Security/IOAuthClientInfo.cs
adrianbanks, line 7: I agree, but this pull request is already large so I don't think it is the right place to do that.

DDDEastAnglia.Tests/Helpers/LoginMethods/LoginMethodIconProviderTests.cs
alastairs: I agree that doesn't feel right, but the reality is:
- We make a change
- ...
- It gets pushed to production
- The site is recompiled on deployment
- The changes take effect

So I don't think the "changing it without recompiling" argument holds much sway here. Perhaps the issue is more one of incorrect decomposition, i.e. the components are at the wrong granularity and so the logic is in the wrong place?

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
alastairs, line 7: `.loginMethod` looks to describe it very well, given that's what the loop variable is called.

DDDEastAnglia/Views/Account/_ExternalLoginsPartial.cshtml
alastairs, line 1: Oh, that makes me sad: they're just lists of login methods, where the individual login methods are rendered slightly differently (links or buttons). Again, I think this points to incorrect decomposition of the partials.

DDDEastAnglia/Helpers/Security/IOAuthClientInfo.cs
alastairs, line 7: I don't agree - the interface was introduced in this pull request, so redefining it to be a number of smaller, more cohesive interfaces, is within the scope of the same pull request.

adrianbanks:
Obsolete now as we've removed the external logins.

PR #331, opened 2015-04-15
Allow sponsors to be added via the website

adrianbanks:
@CraftyFella is continuing the work to allow the sponsors page and sidebar to pick them up from the database. This pull request deals with the admin area only.

### DIFF
--- a/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
+++ b/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
@@ -1,0 +1,23 @@
+﻿using FluentMigrator;
+
+namespace DDDEastAnglia.DatabaseMigrations.Migrations
+{
+    [Migration(20150411)]
+    public class AddSponsorsTable : Migration
+    {
+        public override void Up()
+        {
+            Create.Table("Sponsors")
+                .WithColumn("SponsorId").AsInt32().NotNullable().PrimaryKey()
+                .WithColumn("Name").AsString(50).NotNullable()
+                .WithColumn("Url").AsString(255).NotNullable()
+                .WithColumn("SponsorshipAmount").AsInt32().Nullable()
+                .WithColumn("Logo").AsBinary().NotNullable();
+        }
+
+        public override void Down()
+        {
+            Delete.Table("Sponsors");
+        }
+    }
+}
--- a/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
+++ b/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
@@ -8,7 +8,7 @@ namespace DDDEastAnglia.DatabaseMigrations.Migrations
         public override void Up()
         {
             Create.Table("Sponsors")
-                .WithColumn("SponsorId").AsInt32().NotNullable().PrimaryKey()
+                .WithColumn("SponsorId").AsInt32().NotNullable().Identity().PrimaryKey()
                 .WithColumn("Name").AsString(50).NotNullable()
                 .WithColumn("Url").AsString(255).NotNullable()
                 .WithColumn("SponsorshipAmount").AsInt32().Nullable()
--- a/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
+++ b/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
@@ -12,7 +12,8 @@ namespace DDDEastAnglia.DatabaseMigrations.Migrations
                 .WithColumn("Name").AsString(50).NotNullable()
                 .WithColumn("Url").AsString(255).NotNullable()
                 .WithColumn("SponsorshipAmount").AsInt32().Nullable()
-                .WithColumn("Logo").AsBinary().NotNullable();
+                .WithColumn("Logo").AsBinary(int.MaxValue).NotNullable()
+                .WithColumn("PaymentDate").AsDateTime().Nullable();
         }
 
         public override void Down()
--- a/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
+++ b/DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
@@ -13,7 +13,8 @@ namespace DDDEastAnglia.DatabaseMigrations.Migrations
                 .WithColumn("Url").AsString(255).NotNullable()
                 .WithColumn("SponsorshipAmount").AsInt32().Nullable()
                 .WithColumn("Logo").AsBinary(int.MaxValue).NotNullable()
-                .WithColumn("PaymentDate").AsDateTime().Nullable();
+                .WithColumn("PaymentDate").AsDateTime().Nullable()
+                .WithColumn("ShowPublicly").AsBoolean().NotNullable();
         }
 
         public override void Down()
--- a/DDDEastAnglia.Tests/Helpers/ImageResizeExtensionsTests.cs
+++ b/DDDEastAnglia.Tests/Helpers/ImageResizeExtensionsTests.cs
@@ -8,49 +8,49 @@ namespace DDDEastAnglia.Tests.Helpers
     public sealed class ImageResizeExtensionsTests
     {
         [Test]
-        public void ContrainToHeight_DoesNotResizeTheSuppliedImage_WhenItIsShorterThanTheDesiredHeight()
+        public void ConstrainToHeight_DoesNotResizeTheSuppliedImage_WhenItIsShorterThanTheDesiredHeight()
         {
             const int originalHeight = 300;
 
             using (var bitmap = new Bitmap(200, originalHeight))
             {
                 const int newHeight = 400;
-                var resizedBitmap = bitmap.ContrainToHeightOf(newHeight);
+                var resizedBitmap = bitmap.ConstrainToHeightOf(newHeight);
                 Assert.That(resizedBitmap.Height, Is.EqualTo(originalHeight));
             }
         }
 
         [Test]
-        public void ContrainToHeight_ResizesTheSuppliedImage_WhenItIsTallerThanTheDesiredHeight()
+        public void ConstrainToHeight_ResizesTheSuppliedImage_WhenItIsTallerThanTheDesiredHeight()
         {
             using (var bitmap = new Bitmap(200, 300))
             {
                 const int newHeight = 20;
-                var resizedBitmap = bitmap.ContrainToHeightOf(newHeight);
+                var resizedBitmap = bitmap.ConstrainToHeightOf(newHeight);
                 Assert.That(resizedBitmap.Height, Is.EqualTo(newHeight));
             }
         }
 
         [Test]
-        public void ContrainToWidth_DoesNotResizeTheSuppliedImage_WhenItIsNarrowerThanTheDesiredWidth()
+        public void ConstrainToWidth_DoesNotResizeTheSuppliedImage_WhenItIsNarrowerThanTheDesiredWidth()
         {
             const int originalWidth = 200;
 
             using (var bitmap = new Bitmap(originalWidth, 300))
             {
                 const int newWidth = 400;
-                var resizedBitmap = bitmap.ContrainToWidthOf(newWidth);
+                var resizedBitmap = bitmap.ConstrainToWidthOf(newWidth);
                 Assert.That(resizedBitmap.Width, Is.EqualTo(originalWidth));
             }
         }
 
         [Test]
-        public void ContrainToWidth_ResizesTheSuppliedImage_WhenItIsWiderThanTheDesiredWidth()
+        public void ConstrainToWidth_ResizesTheSuppliedImage_WhenItIsWiderThanTheDesiredWidth()
         {
             using (var bitmap = new Bitmap(200, 300))
             {
                 const int newWidth = 20;
-                var resizedBitmap = bitmap.ContrainToWidthOf(newWidth);
+                var resizedBitmap = bitmap.ConstrainToWidthOf(newWidth);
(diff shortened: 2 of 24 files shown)


DDDEastAnglia.DatabaseMigrations/Migrations/3-AddSponsorsTable.cs
```diff
@@ -0,0 +1,25 @@
...
+                .WithColumn("Url").AsString(255).NotNullable()
+                .WithColumn("SponsorshipAmount").AsInt32().Nullable()
+                .WithColumn("Logo").AsBinary(int.MaxValue).NotNullable()
+                .WithColumn("PaymentDate").AsDateTime().Nullable()
+                .WithColumn("ShowPublicly").AsBoolean().NotNullable();
+        }
+
+        public override void Down()
+        {
+            Delete.Table("Sponsors");
```
alastairs, line 22: This might cause data loss, at worst in production if we have to roll back to the previous migration; at best when switching branches in dev/staging. Are we happy to accept this risk?

adrianbanks, line 22: I think so, as if we were to roll back there is no point of having the sponsor details in the database as they are of no use.

DDDEastAnglia.Tests/Helpers/ImageResizeExtensionsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using DDDEastAnglia.Helpers;
+using NUnit.Framework;
+
+namespace DDDEastAnglia.Tests.Helpers
+{
+    [TestFixture]
+    public sealed class ImageResizeExtensionsTests
+    {
+        [Test]
+        public void ContrainToHeight_DoesNotResizeTheSuppliedImage_WhenItIsShorterThanTheDesiredHeight()
```
alastairs: Typo in these tests: Contrain => Constrain

DDDEastAnglia.Tests/Helpers/ImageResizeExtensionsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            const int originalHeight = 300;
+
+            using (var bitmap = new Bitmap(200, originalHeight))
+            {
+                const int newHeight = 400;
+                var resizedBitmap = bitmap.ContrainToHeightOf(newHeight);
```
alastairs: Typo in the extension method too.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,120 @@
...
+using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.Areas.Admin.Controllers
+{
+    public class SponsorController : Controller
+    {
+        private readonly ISponsorRepository sponsorRepository;
+        private readonly ISponsorSorter sponsorSorter;
+
+        public SponsorController(ISponsorRepository sponsorRepository, ISponsorSorter sponsorSorter)
```
alastairs, line 20: These dependencies should be wrapped in a service that can encapsulate the implementation details of retrieving the sponsors and sorting the list. The responsibility of a `Controller` is to process the web request and hand it off to an application service that can fulfil the request. 

This will help avoid the situation we have in, e.g., the `SessionController`, which has the world injected into it.

adrianbanks, line 20: I originally tried that, but this controller still needs the repository (for the create and delete methods that don't care about sorting), so it seems a bit pointless.

DDDEastAnglia/Areas/Admin/Models/SponsorModel.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+using System.Web;
+
+namespace DDDEastAnglia.Areas.Admin.Models
+{
+    public class SponsorModel
```
alastairs, line 8: This is just called `Sponsor` in the main site. Should we name this the same for consistency across the codebase and with the existing models?

adrianbanks, line 8: The other models in the admin area are all called `*Model`. I'd prefer to keep them consistent within the area.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,120 @@
...
+            return RedirectToAction("Index");
+        }
+
+        public ActionResult Logo(int sponsorId)
+        {
+            var sponsor = sponsorRepository.GetSponsor(sponsorId);
+            return File(sponsor.Logo, "image/png");
+        }
+
+        private SponsorModel CreateSponsorModel(Sponsor sponsor)
```
alastairs, line 77: Should we really be mapping between different representations of the same model? This doesn't look equivalent to mapping across ports/layers (e.g. to persist to a DB); we're mapping between the user-facing site's model and the admin site's model.

alastairs, line 77: In fact, given the models look basically the same (based on these two factory methods), it looks like a duplicated concern.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        [HttpPost]
+        public ActionResult Create(SponsorModel sponsor)
+        {
+            sponsorRepository.AddSponsor(CreateSponsor(sponsor));
+            return RedirectToAction("Index");
+        }
+
+        public ActionResult Delete(int id)
```
alastairs: Which HTTP method is used to invoke this action?

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
adrianbanks, line 77: `SponsorModel` is more like a view model as it is used within the pages. `Sponsor` is the data access class. There is a subtle difference between the two. The view model doesn't contain the logo, but does contain the sponsor's id so that the logo can be loaded.

I agree that it's a bit smelly, but the two are needed so that the validation can be done on creation.

Any ideas on how to make it better?

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,120 @@
...
+                Name = sponsorModel.Name,
+                Url = sponsorModel.Url,
+                SponsorshipAmount = sponsorModel.SponsorshipAmount,
+                Logo = GetLogo(sponsorModel.Logo),
+                PaymentDate = sponsorModel.PaymentDate,
+                ShowPublicly = sponsorModel.ShowPublicly
+            };
+        }
+
+        private byte[] GetLogo(HttpPostedFileBase file)
```
alastairs, line 103: This isn't a `Controller` concern, it's a concern of either the Sponsor service, or an Image infrastructure service. If moved out of the controller, it should take a `Stream`, not an `HttpPostedFileBase`; you should be able to do the same argument checks on a `Stream`.

([Jimmy Bogard's blog post](https://lostechies.com/jimmybogard/2008/08/21/services-in-domain-driven-design/) gives a good overview of the different types of service: application, domain and infrastructure.)

DDDEastAnglia/Areas/Admin/Views/Sponsor/Create.cshtml
```diff
@@ -0,0 +1,71 @@
...
+        <div class="controls">
+            @Html.TextBoxFor(m => Model.PaymentDate, new { type = "date" })
+            @Html.ValidationMessageFor(m => Model.PaymentDate)
+        </div>
+    </div>
+
+    <div class="control-group">
+        @Html.Label("Logo", new { @class = "control-label" })
+        <div class="controls">
+            <input type="file" name="Logo" data-val="true" data-val-required="A logo must be supplied." />
```
alastairs: The `required` attribute would be a useful addition here too: it will cause the browser to prompt the user to include a logo. Other fields would benefit from the same.

DDDEastAnglia/Areas/Admin/Views/Sponsor/Delete.cshtml
```diff
@@ -0,0 +1,18 @@
+﻿@model DDDEastAnglia.Models.Sponsor
+
+@{
+    ViewBag.Title = string.Format("Delete {0}?", Model.Name);
+}
+
+<h2>@ViewBag.Title</h2>
+
+<p>Deleting a sponsor cannot be undone. Are you <em>really</em> sure you want to delete this sponsor?</p>
```
alastairs, line 9: This feels like it should be reusable beyond this scenario (and don't we have other deletion confirmations elsewhere in the system?).

DDDEastAnglia/Areas/Admin/Views/Sponsor/Create.cshtml
adrianbanks: The client side validation is already done by MVC based on the markup in the view model. That all works except for this logo, as file attributes don't work with it, so you have to add the `data-val` attributes manually.

DDDEastAnglia/Areas/Admin/Views/Sponsor/Delete.cshtml
adrianbanks, line 9: I agree - there was several bits in making the CRUD for sponsors that were very similar to those for other object types. Something to be done later though, otherwise this pull request will get huge.

DDDEastAnglia/DataAccess/SimpleData/SponsorRepository.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+using DDDEastAnglia.Models;
+using Simple.Data;
+
+namespace DDDEastAnglia.DataAccess.SimpleData
+{
+    public class SponsorRepository : ISponsorRepository
```
alastairs, line 8: Can we go with query objects rather than repositories?

adrianbanks, line 8: How about we convert them all over in one go in a separate change?

DDDEastAnglia/DataAccess/SponsorSorter.cs
```diff
@@ -0,0 +1,44 @@
...
+using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.DataAccess
+{
+    public interface ISponsorSorter
+    {
+        IEnumerable<Sponsor> Sort(IEnumerable<Sponsor> sponsors);
+    }
+
+    public class SponsorSorter : ISponsorSorter
```
alastairs: Is this a `DescendingAmountSponsorSorter`? Or something similar? Perhaps `DefaultSponsorSorter` would be sufficient. Depends how likely we are to want to tweak the algorithm, etc.

adrianbanks: It is probably a default one, as I can't see why we would want to sort them in a different order.

DDDEastAnglia/Helpers/ImageResizeExtensions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Drawing;
+
+namespace DDDEastAnglia.Helpers
+{
+    public static class ImageResizeExtensions
```
alastairs, line 6: Typos in both extension method names.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
alastairs, line 20: Those would be features of the service(s) too. The point of the service is to provide the domain operations for the system (such as adding a sponsor)*. The repository is an implementation detail that shouldn't be exposed to the controller.
- I think technically this would be an application service, because we're starting out with a non-domain representation of the concern, i.e. a ViewModel.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
alastairs, line 77: Moving `Sponsor` to the `DataAccess` namespace would be fine. It was just that, being in the `Models` namespace it looked like they were both ViewModels.

DDDEastAnglia/Areas/Admin/Models/SponsorModel.cs
alastairs, line 8: :+1: See my other comment about confusion as to the intended purpose of `Sponsor`/`SponsorModel`.

DDDEastAnglia/Areas/Admin/Views/Sponsor/Create.cshtml
alastairs: jQuery validation will work with the HTML5 validation attributes so you don't have to use the custom data-\* attributes. These attributes have the added bonus that if Javascript is unavailable for any reason the validation still happens.

DDDEastAnglia/DataAccess/SimpleData/SponsorRepository.cs
alastairs, line 8: I tried that already and found that delivering in small incremental chunks is the way to go :smile: 

More seriously, I'd really rather we didn't introduce new repositories and migrated existing ones over to query objects as we touch them. Otherwise it will be an ever-increasing problem that we won't be able to resolve.

DDDEastAnglia/Helpers/ImageResizeExtensions.cs
adrianbanks, line 6: No, already corrected them before you commented.

CraftyFella:
I've create a new pull request..mainly because I don't know what I'm doing with this stuff.. which has all the above changes pull the stuff @JonCanning and I worked on.. which is using the above for the sidebar and sponsors page..

PR #332, opened 2015-04-28
Show sponsors that have been added via the admin page

CraftyFella:
This follows on from @adrianbanks #331 pull request.

Basically it uses the sponsors from the admin page and shows them on the side bar and the sponsors page.. 

All the existing comments in #331 still apply.. 

All feedback most welcome indeed.

### DIFF
--- a/DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
+++ b/DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
@@ -24,9 +24,9 @@ namespace DDDEastAnglia.Areas.Admin.Controllers
                 throw new ArgumentNullException("sponsorRepository");
             }
 
-            if (sponsorRepository == null)
+            if (defaultSponsorSorter == null)
             {
-                throw new ArgumentNullException("sponsorRepository");
+                throw new ArgumentNullException("defaultSponsorSorter");
             }
 
             this.sponsorRepository = sponsorRepository;


DDDEastAnglia.Tests/Controllers/HomeControllerBuilder.cs
```diff
@@ -0,0 +1,28 @@
+using DDDEastAnglia.Controllers;
+using DDDEastAnglia.DataAccess;
+
+namespace DDDEastAnglia.Tests.Controllers
+{
+    public class HomeControllerBuilder
+    {
+        private IConferenceLoader conferenceLoader = new ConferenceLoaderBuilder().Build();
```
alastairs: I assume `ConferenceLoaderBuilder` creates a real implementation of `IConferenceLoader` and not a mocked one? I think for the default values here, I'd rather not use the Builder, so you can see from reading the class what you get by default.

DDDEastAnglia.Tests/Controllers/HomeControllerTests.cs
```diff
@@ -14,8 +13,10 @@ public void SubmitSessionLinks_ShouldBeHidden_WhenSessionSubmissionIsNotOpen()
                                         .WithSessionSubmissionClosed()
                                         .Build();
 
-            var model = new HomeController(conferenceLoader)
-                            .About().GetViewModel<AboutViewModel>();
+            var model = new HomeControllerBuilder()
```
alastairs: I'd probably pull the controller out into a variable now: the `Build()` call in the middle looks a bit odd.

DDDEastAnglia.Tests/DateExtensions.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+
+namespace DDDEastAnglia.Tests
+{
+    public static class DateExtensions
+    {
+        public static DateTime January(this int day, int year)
```
alastairs: Noice :smile: :+1: 

(Sorry for the triviality, but please could you reformat - blank lines between methods, no blank line after last method?)

DDDEastAnglia.Tests/InMemoryRepository.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace DDDEastAnglia.Tests
+{
+    internal class InMemoryRepository<T>
+    {
+        private readonly Func<T, int> keyFunc;
+        private readonly ConcurrentDictionary<int, T> store = new ConcurrentDictionary<int, T>();
+
+        public InMemoryRepository(Func<T, int> keyFunc)
+        {
+            this.keyFunc = keyFunc;
```
alastairs: This needs a null check.

DDDEastAnglia.Tests/InMemoryRepository.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+
+namespace DDDEastAnglia.Tests
+{
+    internal class InMemoryRepository<T>
+    {
+        private readonly Func<T, int> keyFunc;
+        private readonly ConcurrentDictionary<int, T> store = new ConcurrentDictionary<int, T>();
+
+        public InMemoryRepository(Func<T, int> keyFunc)
```
alastairs: It's not immediately obvious to me what I should pass as the `keyFunc`: generics and primitives aren't that expressive. Can that be improved somehow?

On a related note, some guidance on how to write a key function might be good too? I'm guessing `_ => 1` isn't going to be very useful... but equally, I'm not sure what _would_ be useful.

DDDEastAnglia.Tests/InMemoryRepository.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace DDDEastAnglia.Tests
+{
+    internal class InMemoryRepository<T>
```
alastairs: Can `T` be constrained at all? At least to `class`, perhaps? An `InMemoryRepository<string>` might not be that valuable. (Or perhaps it might?)

DDDEastAnglia.Tests/SponsorBuilder.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private byte[] logo;
+        private string name;
+        private DateTime? paymentDate;
+        private bool showPublicly = true;
+        private int sponsorId;
+        private int sponsorshipAmount;
+        private string url;
+
+        public Sponsor Build()
```
alastairs: I tend to put this at the bottom, as it's the least interesting bit of a Builder (to me). 

I'm also not convinced a Builder brings much valuable to a publicly-mutable type like `Sponsor`, unless it's doing a bit more than simply setting the properties on the type. (Mark Seemann terms Builder methods like these "higher-order Builder methods"; I'm not sure that's technically correct, but let's roll with it.)

Higher-order Builder methods for Sponsor might be something like `AtPlatinumLevel()` - something more related to our domain than just what they've paid and when.

DDDEastAnglia.Tests/SponsorModelQueryBuilder.cs
```diff
@@ -0,0 +1,15 @@
+using DDDEastAnglia.Controllers;
+using DDDEastAnglia.DataAccess;
+
+namespace DDDEastAnglia.Tests
+{
+    public class SponsorModelQueryBuilder
```
alastairs: Again, is this buying us much or could we make do with a simpler [SUT Factory](http://blog.ploeh.dk/2009/02/13/SUTFactory/) for now?

alastairs: In fact, is it used? The `sponsor_list.context` appears to set this up manually.

DDDEastAnglia.Tests/Sponsors/sponsor_list/context.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using DDDEastAnglia.Controllers;
+using DDDEastAnglia.DataAccess;
+using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.Tests.Sponsors.sponsor_list
+{
+    public class context
```
alastairs: Fixture Objects. Nice. :+1:

DDDEastAnglia.Tests/Sponsors/sponsor_list/Sponsors_who_havent_paid_yet.cs
```diff
@@ -0,0 +1,31 @@
+using System.Linq;
+using NUnit.Framework;
+
+namespace DDDEastAnglia.Tests.Sponsors.sponsor_list
+{
+    [TestFixture]
+    public sealed class Sponsors_who_havent_paid_yet : context
+    {
+        public Sponsors_who_havent_paid_yet()
```
alastairs: Question: should we make this a method and decorate it with `[TestFixtureSetUp]`?

The reason I ask is because I wasn't expecting to see this pattern. In a suite of Xunit-based tests, I think I'd have more quickly recognised this as being a constructor, but because it's an NUnit suite I was surprised not to see `[Test]` on this method. 

Not sure whether this is me being blind/stupid or a [POLA](http://en.wikipedia.org/wiki/Principle_of_least_astonishment) thing.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,120 @@
...
+using DDDEastAnglia.DataAccess.SimpleData.Models;
+using DDDEastAnglia.Helpers;
+using SponsorModel = DDDEastAnglia.Areas.Admin.Models.SponsorModel;
+
+namespace DDDEastAnglia.Areas.Admin.Controllers
+{
+    public class SponsorController : Controller
+    {
+        private readonly ISponsorRepository sponsorRepository;
+        private readonly DefaultSponsorSorter _defaultSponsorSorter;
```
alastairs: Fields shouldn't be prefixed with an underscore.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public SponsorController(ISponsorRepository sponsorRepository, DefaultSponsorSorter _defaultSponsorSorter)
```
alastairs: [It looks like we used to pass in an `ISponsorSorter`](https://github.com/adrianbanks/dddeastanglia/compare/add-sponsors-to-db...CraftyFella:pull-sponsors-for-sidebar#diff-67f297aaf446858ad772daa1dafa8025L20): why are we now passing the concrete type?

DDDEastAnglia/Controllers/HomeController.cs
```diff
@@ -9,15 +9,17 @@ namespace DDDEastAnglia.Controllers
     public class HomeController : Controller
     {
         private readonly IConferenceLoader conferenceLoader;
+        private readonly SponsorModelQuery sponsorModelQuery;
 
-        public HomeController(IConferenceLoader conferenceLoader)
+        public HomeController(IConferenceLoader conferenceLoader, SponsorModelQuery sponsorModelQuery)
```
alastairs: I feel as though there should be an interface that this implements - `IQuery<T>`, perhaps? Maybe this is something to add when we have more than one query object, though :smile:

It also feels as though we're injecting the wrong thing into this controller: a query object is a data access concern, so shouldn't that be used from a service rather than a controller action? I like controller actions to be featherweight and only responsible for doing the webby bits before handing off to/returning from a service method.

alastairs: It also needs a null check.

DDDEastAnglia/Controllers/SponsorController.cs
```diff
@@ -0,0 +1,28 @@
...
+using DDDEastAnglia.Domain;
+
+namespace DDDEastAnglia.Controllers
+{
+    public class SponsorController : Controller
+    {
+        private readonly SponsorModelQuery sponsorModelQuery;
+        private readonly SponsorLogoService sponsorLogoService;
+
+        public SponsorController(SponsorModelQuery sponsorModelQuery, SponsorLogoService sponsorLogoService)
```
alastairs: Null checks. Interfaces?

DDDEastAnglia/Controllers/SponsorLogoService.cs
```diff
@@ -0,0 +1,21 @@
+using System.Web.Mvc;
+using DDDEastAnglia.DataAccess;
+
+namespace DDDEastAnglia.Controllers
+{
+    public class SponsorLogoService
+    {
+        private readonly ISponsorRepository sponsorRepository;
+
+        public SponsorLogoService(ISponsorRepository sponsorRepository)
```
alastairs: Null check.

DDDEastAnglia/Controllers/SponsorLogoService.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            this.sponsorRepository = sponsorRepository;
+        }
+
+        public FileContentResult Get(int sponsorId)
```
alastairs: A service shouldn't return a framework type, that should be for the Controller to decide.

DDDEastAnglia/Controllers/SponsorModelQuery.cs
```diff
@@ -0,0 +1,40 @@
...
+using DDDEastAnglia.DataAccess.SimpleData.Models;
+using DDDEastAnglia.Models;
+
+namespace DDDEastAnglia.Controllers
+{
+    public class SponsorModelQuery
+    {
+        private readonly ISponsorRepository sponsorRepository;
+
+        public SponsorModelQuery(ISponsorRepository sponsorRepository)
```
alastairs: Null check.

alastairs: So we have a query object wrapping a repository - that feels like a layer of abstraction too many. I'd like us to move to query objects (and introduce query objects for new work we do), but I don't see any value in using them to wrap a repository. My understanding of them is that they're more like SRP-compliant repositories: they talk directly to the ORM/data access client, but only do One Thing™ rather than All The Data Accesses.

DDDEastAnglia/DataAccess/DefaultSponsorSorter.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    public static class SponsorExtensions
+    {
+        public static IEnumerable<Sponsor> OrderBySponsorSorter(this IEnumerable<Sponsor> sponsors)
+        {
+            return new DefaultSponsorSorter().Sort(sponsors);
+        }
+    }
+
+    public class DefaultSponsorSorter : ISponsorSorter
```
alastairs: [This was sealed](https://github.com/adrianbanks/dddeastanglia/compare/add-sponsors-to-db...CraftyFella:pull-sponsors-for-sidebar#diff-f010e181923c8492ec8d18b90389e2f1L9), but is no longer. Was that intentional?

DDDEastAnglia/DataAccess/DefaultSponsorSorter.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using DDDEastAnglia.DataAccess.SimpleData.Models;
+
+namespace DDDEastAnglia.DataAccess
+{
+    public static class SponsorExtensions
+    {
+        public static IEnumerable<Sponsor> OrderBySponsorSorter(this IEnumerable<Sponsor> sponsors)
```
alastairs: The method name implies that it could take different `SponsorSorter`s, but it does not; as a result, it's not very Open-Closed. I think I'd also prefer the `SponsorSorter` be given to the `SponsorService` (or whatever) so that it is able to provide a sorted list of sponsors, on a flexible basis, but that's possibly a stylistic thing.

DDDEastAnglia/Models/SponsorModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace DDDEastAnglia.Models
+{
+    public class SponsorModel
```
alastairs: I think we now have three classes that represent this information. Are they all sufficiently distinct to warrant their separate existences?

DDDEastAnglia/Views/Shared/_Layout.cshtml
```diff
@@ -1,107 +1,108 @@
...
+                                Cambridge<br />
+                                CB3 9DP
+                            </p>
+                        </section>
 
-                            @Html.Partial("_Sponsors")
-                        </div>
+                        @Html.Action("Details", "Timeline")
+                        @{
+                            Html.RenderAction("Sidebar", "Sponsor");
```
alastairs, line 65: I think this is the only change in this file we actually care about that much. If you can, please try to keep functional changes separate from formatting changes.

Why does this call have to be in braces, but the line above does not?

DDDEastAnglia/Views/Shared/_Sponsors.cshtml
```diff
@@ -1,63 +1,11 @@
-﻿<section id="sponsors">
+﻿@model IEnumerable<DDDEastAnglia.Models.SponsorModel>
+<section id="sponsors">
```
alastairs, line 2: Can the two sponsors' views now be properly merged, so we get actual reuse of code? :smile:

CraftyFella:
Awesome stuff @alastairs I'll go through your comments later tonight. :)

DDDEastAnglia.Tests/Controllers/HomeControllerBuilder.cs
CraftyFella: Justification for this:

It uses the builders to create the conferenceloader so that when someone changes the dependencies of the conferenceloader you don't have loads of constructors to fix.

This is what happened to me when I added a dependency to the HomeController.. I had 15 constructors to go and fix.. 

Not too fussed though as I imagine conferenceloader won't change too much

DDDEastAnglia.Tests/Controllers/HomeControllerTests.cs
CraftyFella: I'm happy to loose the builder all together if you want and replace it with a simple method in the class...

DDDEastAnglia.Tests/DateExtensions.cs
CraftyFella: No probs..will do

DDDEastAnglia.Tests/InMemoryRepository.cs
CraftyFella: Do the null checks actually help? We are using Ninject which will blow up when it can't find a dependency.. so the null checks will never throw?

DDDEastAnglia.Tests/InMemoryRepository.cs
CraftyFella: Mmm good question.. not sure.. I didn't want to introduce a load of boiler plate classes for this so a simple delegate seemed more appropriate.. I could have a named delegate? or use an alias? 

It's used to tell the generic class where the id is. .for get by id calls. 

As we're using SQL.. our primary keys will be ints I'm assuming.

how about `primarykeyfinder`?

usage is 

`(x => x.Id)`

DDDEastAnglia.Tests/InMemoryRepository.cs
CraftyFella: good call there.. adding a class constraint makes sense..

DDDEastAnglia.Tests/SponsorBuilder.cs
CraftyFella: I like the idea of the AtPlatinumLevel() idea.. very nice.

Justification:
For me a builders job is simple.. create a valid object every time.. with default values.. 

Example being is `new Sponsor()` is that valid? I don't want to have to this about it.. so I'll use a builder which will always create me a valid one.

DDDEastAnglia.Tests/SponsorModelQueryBuilder.cs
CraftyFella: I'll check that out.. thanks

DDDEastAnglia.Tests/Sponsors/sponsor_list/context.cs
CraftyFella: didn't know they had a name.. it's my new default for testing.. glad you like it.

DDDEastAnglia.Tests/Sponsors/sponsor_list/Sponsors_who_havent_paid_yet.cs
CraftyFella: the behaviour in nunit is different.. a constructor acts like the `[TestFixtureSetup]` but I think in xunit the constructor acts more like `[Setup]`

I don't mind.. if it adds more clarity then I'm all for it..

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
CraftyFella: Oops :tongue:

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
CraftyFella: I had a discussion with @adrianbanks about this.. 

IMHO the `DefaultSponsorSorter` doesn't need an interface.. it's just behaviour.. in fact it could be a pure function.. as it takes an `IEnumerable<T>` and returns an `IEnumerable<T>`.. and doesn't have any dependencie so passing in a delegate would allow it to be changed?

DDDEastAnglia/Controllers/HomeController.cs
CraftyFella: Ahh no probs.. was trying to re-use what we'd already done.. 

Have you looked at the existing IQuery? Is that what you're suggesting we use?

Question.. 

What should the query return? ViewModels DomainModels, Storage Models?

I'm also confused on what we're trying to do here.. 

Perhaps an example of something in the existing code base which does this correctly would help? Is there one?

Any hints most welcome :+1:

DDDEastAnglia/Controllers/SponsorController.cs
CraftyFella: See above comment about this.

DDDEastAnglia/Controllers/SponsorLogoService.cs
CraftyFella: That makes sense. :+1:

DDDEastAnglia/Controllers/SponsorModelQuery.cs
CraftyFella: Cool.. do we have any that use simple data in the current code base? :+1:

DDDEastAnglia/DataAccess/DefaultSponsorSorter.cs
CraftyFella: nope.. must have been a merge problem

DDDEastAnglia/DataAccess/DefaultSponsorSorter.cs
CraftyFella: See previous comment about pure function.. but if we go with a class then I like the idea of

`SortUsing<DefaultSponsorSorter>()`

or if we do inject it.. then

`SortUsing(injectedSorter)`

Any preference?

DDDEastAnglia/Models/SponsorModel.cs
CraftyFella: I'd like some guidence on how we pass data from UI to DB and back again..

What are the models we use etc..

DDDEastAnglia/Views/Shared/_Layout.cshtml
CraftyFella, line 65: Good call.. :+1:  I'll try and fix this.

DDDEastAnglia/Views/Shared/_Sponsors.cshtml
CraftyFella, line 2: There were subtle differences in the markup.. I think with some css foo it should be possible.

DDDEastAnglia.Tests/InMemoryRepository.cs
CraftyFella: Wait.. this isn't in in the app.. just the tests.. so null check makes sense..

DDDEastAnglia.Tests/SponsorModelQueryBuilder.cs
CraftyFella: At the moment it's only used in the `HomeController` so If I replace the `HomeControllerBuilder` with a simple method then all this goes away for now..

DDDEastAnglia.Tests/DateExtensions.cs
adrianbanks: Instead of this class, we could pull in [Humanizer](https://github.com/MehdiK/Humanizer) via NuGet.

DDDEastAnglia.Tests/SponsorBuilder.cs
adrianbanks: > I like the idea of the AtPlatinumLevel() idea.. very nice.

I like the idea too, but wouldn't we need to change the tests when we change the sponsorship amounts? Also, the tests (and the code) don't really care about what level someone has sponsored us at, just how much they have sponsored.

DDDEastAnglia/DataAccess/DefaultSponsorSorter.cs
adrianbanks: I don't like this extension method. There is already an `OrderBy` and `OrderByDescending` methods in linq that take a comparer so this extension method feels unnecessary.

DDDEastAnglia/Views/Home/Sponsors.cshtml
```diff
@@ -1,68 +1,20 @@
...
-
-<div class="sponsor span5">
-    <a href="http://www.pluralsight.com"><img src="@Url.Content("~/Images/sponsors/pluralsight.png")" /></a>
-</div>
-
-<div class="sponsor span5">
-    <a href="http://www.skillsmatter.com"><img src="@Url.Content("~/Images/sponsors/skillsmatter.png")" /></a>
-</div>
+@foreach (var sponsor in Model)
+{<div class="sponsor span5">
```
adrianbanks: You need a line break here to put the `div` on a new line.

DDDEastAnglia/Views/Shared/_Sponsors.cshtml
```diff
@@ -1,63 +1,11 @@
...
-    </div>
-
-    <div class="sponsor">
-        <a href="http://www.skillsmatter.com"><img src="@Url.Content("~/Images/sponsors/skillsmatter.png")" /></a>
-    </div>
-
-    <div class="sponsor">
-        <a href="http://www.ecmselection.co.uk"><img src="@Url.Content("~/Images/sponsors/ecmselection.png")" /></a>
+    @foreach (var sponsor in Model)
+    {<div class="sponsor">
```
adrianbanks: You need a line break here as well.

adrianbanks:
You have a few classes and namespaces that are lowercase (all in the `sponsor_list` folder). Could you fix them to make them match the convention used in the rest of the code?

DDDEastAnglia/Views/Shared/_Layout.cshtml
CraftyFella, line 65: Ok.. I remember now.. the line above is rendering a link.. where as the RenderAction is actually because we're calling the Sidebar action in the Sponsor Controller.. which requires access to the SponsorQuery.. Hope that makes sense with the Html.RenderAction.

see http://www.dotnet-tricks.com/Tutorial/mvc/Q8V2130113-RenderPartial-vs-RenderAction-vs-Partial-vs-Action-in-MVC-Razor.html

CraftyFella:
@adrianbanks sure.. no probs

CraftyFella:
@adrianbanks Oops I misread @alastairs comment as we want underscores for fields.. 

Which makes alot of sense IMHO.. I will change my settings and push a new commit...

DDDEastAnglia/Views/Home/Sponsors.cshtml
CraftyFella: as in `<br>` ?

adrianbanks: No, just a CRLF to put the `div` tag on the next line (plus indented correctly).

DDDEastAnglia.Tests/DateExtensions.cs
CraftyFella: @adrianbanks  Doesn't look like it supports the same syntax to me.

https://dotnetfiddle.net/jKiLVY

`In.FebruaryOf(2009) // Returns 1st of February of 2009`

vs

`1.February(2009)`

doesn't look like you can do 

`2.February(2009)`

adrianbanks: You can do `On.February.The1st.In(2009)`.

CraftyFella: still prefer mine :tongue: but why the hell not.

adrianbanks: Don't mind either way, but thought it would be good to not reinvent the wheel.

CraftyFella:
Ok so that will do for tonight.. keep the comments coming. it's been a great experience.. 

I'm waiting for feedback on my comments to @alastairs until I do any more

DDDEastAnglia.Tests/InMemoryRepository.cs
alastairs: `idSelector` perhaps?

DDDEastAnglia.Tests/SponsorBuilder.cs
alastairs: There are two things here:
1. The amount sponsored is primarily (but not exclusively) a function of the sponsorship level (Platinum/Gold/Silver/Bronze)
2. We _always_ talk about sponsorship levels rather than amounts, so the code is modelling an implementation detail of the domain, rather than expressing the concept in domain terminology.

DDDEastAnglia.Tests/Sponsors/sponsor_list/Sponsors_who_havent_paid_yet.cs
alastairs: Yes, that's a fair point: Xunit gives you an instance per test (one of the reasons it's better than NUnit 😆)

Second opinion, @adrianbanks?

DDDEastAnglia.Tests/SponsorBuilder.cs
adrianbanks: That's true, but in reality over the past two years we haven't been able to group the sponsors that neatly into those amounts. Even though we have the levels, many sponsors don't fit it that closely.

DDDEastAnglia/Areas/Admin/Controllers/SponsorController.cs
alastairs: My concern is that the class is no longer open for extension by changing this to a concrete type. I'm not fussed exactly how we achieve that - delegates are fine, although I quite like the Strategy pattern approach because it allows you to put names on the concepts. It's very much an OO vs. Functional thing, though, and given C# straddles the two paradigms, there's no clear answer.

I tend to use delegates (Functional style) when the delegate will nearly always be a trivial lambda, and the Strategy pattern when the situation is more complex. The SponsorSorter concern may be still sufficiently complex to justify a full-on Strategy implementation.

DDDEastAnglia/Controllers/HomeController.cs
alastairs: > Perhaps an example of something in the existing code base which does this correctly would help? Is there one?

To my knowledge, there isn't 😔 I'd like the interface to return Domain models: everything should deal in those until it has to go to an external system of some form (so, e.g., a query object maps data access results to a Domain model).

DDDEastAnglia/Controllers/SponsorModelQuery.cs
alastairs: Not in production, I don't think.  I have a branch in my fork called something like  `repositories-to-query-objects` that might provide some inspiration...

DDDEastAnglia.Tests/InMemoryRepository.cs
alastairs: IMO, null checks do make sense because they assert the preconditions for using the class hold true. Yes, it's true to say that Ninject will b0rk before we ever get a chance to pass a null to most things, but we do also create a lot of stuff manually in our code base. 

More generally, I don't feel an application should be dependent on the behaviour of a third-party library in this way to be minimally functional. We should be able to remove Ninject and replace it with manual composition (or another IoC container) and still have everything work fine. 

(Additionally, this is one of the few things we actually _are_ consistent about in the code base 😆)

alastairs:
I think I've responded to everything that's outstanding; if I've missed anything, just DM me on Slack (to keep the pull request thread size down).

CraftyFella:
That's excellent.. thanks @alastairs and @adrianbanks for the feedback. Will have another go tonight

alastairs:
For the sake of clarity, here's what I think is left to do:
- [x] Humanizr. I don't much care about this as I prefer @CraftyFella's more concise syntax. @adrianbanks, how much would you prefer to pull in Humanizr over this class of extension methods?
- [x] Null check on `HomeController` constructor parameters.
- [x] Introduce an interface on `SponsorModelQuery`?
- [x] Null checks on `SponsorController` constructor parameters.
- [x] Interface on `SponsorLogoService`.
- [x] Null check on `SponsorLogoService` constructor parameter.
- [x] Null check on `SponsorModelQuery` constructor parameter.
- [ ] Resolve the tension between repositories and query objects: we should use one or the other. We currently have repositories, but I'd like us to move towards query objects. I'm happy for this branch to break with the convention of repositories to use query objects, but the current situation of query objects depending on repositories is wrong and should be removed which ever direction is taken. 
- [x] Replace `OrderBySponsorSorter()` extension method with LINQ's `OrderBy`.

adrianbanks:
Humanizr - wasn't really bothered either way, but pulled it in as it means less code for us to maintain.

PR #334, opened 2015-05-17
Email submissions to speakers

alastairs:
There's quite a lot of refactoring here, but I think we've ended up with quite a nice way of notifying users of events in the system. In theory, we should be able to hook up other transport mechanisms beyond email, such as Twitter, with little extra code. We'll see...

Note that the history has been fairly severely squashed together to make it a little easier to follow. The original history, if you want the detail behind some of the commits, is available [here](https://github.com/dddeastanglia/DDDEastAnglia/compare/staging...alastairs:email-submissions-to-speakers).

Can we get a review and merge, @adrianbanks @CraftyFella @JonCanning?


DDDEastAnglia.Tests/Services/Messengers/Email/SessionSubmissionMessengerFacts.cs
```diff
@@ -0,0 +1,40 @@
...
+using DDDEastAnglia.Models;
+using DDDEastAnglia.Services.Messenger.Email;
+using NSubstitute;
+using NUnit.Framework;
+using MailMessage = DDDEastAnglia.Services.Messenger.Email.MailMessage;
+
+namespace DDDEastAnglia.Tests.Services.Messengers.Email
+{
+    [TestFixture]
+    public class SessionSubmittedMailMessengerFacts
```
alastairs: Should be renamed to `EmailMessengerFacts`.

DDDEastAnglia.Tests/Services/Messengers/Email/Templates/SessionCreated.txt
```diff
@@ -0,0 +1,2 @@
+﻿Title: [SessionTitle]
```
alastairs: This file should be deleted: the email templates are in the Emails.resx file now.

DDDEastAnglia.Tests/Services/Messengers/Email/Templates/SessionUpdated.txt
```diff
@@ -0,0 +1,4 @@
+﻿Updated!
```
alastairs: This file should also be deleted.

DDDEastAnglia.Tests/Controllers/ResetPasswordControllerTests.cs
```diff
@@ -16,45 +18,37 @@ public sealed class ResetPasswordControllerTests
         [Test]
         public void TestThat_Ctor_ThrowsAnException_WhenTheSuppliedContextIsNull()
         {
-            var resetPasswordThingy = Substitute.For<IResetPasswordThingy>();
-            var resetPasswordEmailSender = Substitute.For<IResetPasswordEmailSender>();
-            Assert.Throws<ArgumentNullException>(() => new ResetPasswordController(null, resetPasswordThingy, resetPasswordEmailSender));
+            var resetPasswordThingy = Substitute.For<IResetPasswordService>();
```
alastairs: This variable needs renaming too (plus other instances in this file).

DDDEastAnglia/Controllers/SessionController.cs
```diff
@@ -110,7 +113,13 @@ public ActionResult Create([Bind(Exclude = "Votes")] Session session)
             if (ModelState.IsValid)
             {
                 var addedSession = sessionRepository.AddSession(session);
-                return RedirectToAction("Details", new {id = addedSession.SessionId});
+
+                UserProfile speakerProfile = userProfileRepository.GetUserProfileByUserName(User.Identity.Name);
+                
+                var sessionCreatedMailTemplate = SessionCreatedMailTemplate.Create(addedSession);
+                emailMessengerFactory.CreateEmailMessenger(sessionCreatedMailTemplate).Notify(speakerProfile);
```
alastairs, line 120: This is basically the one thing that I don't like about the design that we have for sending users messages: the `EmailMessenger` has to exist on a per-email basis, and so needs to be called into existence before we can tell it to notify the given user. 

This could be improved by passing the `IMailTemplate` to `Notify()` as well as the speaker profile, so it becomes more like "Notify this user using this pre-determined message template", which makes sense in some respects, but this approach ties the `IMessenger` interface to email messages. A way to fix that would be to make `IMessenger` generic on the thing passed to `Notify()`, but that feels like lazy design to me: there should be a common interface for messages to send to users, no matter what transport mechanism is in use. 

I think maybe I'll try hooking in Twitter notifications (or similar - texts, perhaps) and see what that does to the design of these interfaces.

DDDEastAnglia/Helpers/Email/SendGrid/IRenderer.cs
```diff
@@ -0,0 +1,7 @@
+namespace DDDEastAnglia.Helpers.Email.SendGrid
+{
+    public interface IRenderer
```
alastairs, line 3: This should probably have a better name. Am open to suggestions :smile:

DDDEastAnglia/Helpers/MvcHelperModule.cs
```diff
@@ -17,6 +18,7 @@ public override void Load()
                                     .InNamespaceOf<IControllerInformationProvider>()
                                     .Excluding<AlphabeticalSort>()
                                     .Excluding<RandomSort>()
+                                    .Excluding<HtmlRenderer>()
```
alastairs, line 21: This is bound in `EmailModule` further up in the diff.

alastairs:
This needs some manual testing. I haven't been able to do that because it turns out I haven't yet installed SQL Server on my new dev VM. @philpursglove could you perhaps take it for a quick spin too? As before, creating a session should get as far as making the SendGrid API calls in `SendGridPostman`, but probably crash and burn there because we don't have the SendGrid config details in dev environments.

alastairs:
Manual test succeeded on my machine :smile:

philpursglove:
Github wouldn't let me add a comment saying I was going to test it this afternoon :cry:

alastairs:
...and yet it let you add _that_ comment. Weird. Do give it a test anyway - "works on my machine™" is better if it's "works on mine and Phil's machines™" :wink:

DDDEastAnglia.Tests/ControllerTestExtensions.cs
```diff
@@ -26,5 +28,21 @@ public static void SetupWithHttpContextAndUrlHelper(this Controller controller)
             controller.ControllerContext = controllerContext;
             controller.Url = new UrlHelper(new RequestContext(controller.HttpContext, new RouteData()), RouteTable.Routes);
         }
+
+        public static void SetupWithAuthenticatedUser(this Controller controller, UserProfile user)
+        {
+            if (user == null) return;
```
adrianbanks: Needs braces.

DDDEastAnglia.Tests/Controllers/SessionControllerTests.cs
```diff
@@ -73,29 +69,63 @@ public void CannotBeginToDeleteASession_WhenTheSessionBelongsToAnotherUser()
...
+            var controller = new SessionControllerBuilder().Updating(session).Build();
 
             var result = controller.DeleteConfirmed("fred", 1);
 
             Assert.That(result, Is.InstanceOf<HttpUnauthorizedResult>());
         }
 
-        private static class SessionControllerFactory
+        [TestFixture]
+        public class EmailTests
```
adrianbanks: I find this being a nested class a bit clumsy, especially as we don't get the benefit of the fluent naming.

DDDEastAnglia.Tests/Helpers/Email/SendGrid/HtmlRendererFacts.cs
```diff
@@ -0,0 +1,40 @@
...
+    [TestFixture]
+    public class HtmlRendererFacts
+    {
+        [TestFixture]
+        public class Render_Should
+        {
+            [Test]
+            public void Convert_Markdown_To_HTML()
+            {
+                var sut = CreateSut();
```
adrianbanks: I find calling the variable `sut` and the method `CreateSut()` distracts me when reading the tests. Calling the variable `htmlRenderer` would help greatly, even if the method wasn't called `CreateRenderer()`..

DDDEastAnglia.Tests/Controllers/SessionControllerTests.cs
alastairs: I know what you mean. I did it for explicit categorisation - these tests cover a subset of the behaviour of the `SessionController`. I then felt compelled to follow the conventions in the file, so ended up with the clunky names. Perhaps I'll remove the nested class and just make the two tests top-level tests.

DDDEastAnglia.Tests/Services/Messengers/Email/EmailMessengerFacts.cs
```diff
@@ -0,0 +1,40 @@
...
+                var template = Substitute.For<IMailTemplate>();
+                template.RenderBody().Returns("Message body");
+                template.RenderSubjectLine().Returns("Message subject");
+
+                var messenger = new EmailMessenger(postman, template);
+                messenger.Notify(new UserProfile {EmailAddress = "speaker@dddeastanglia.com"});
+
+                var message = new MailMessage
+                {
+                    From = new MailAddress("admin@dddeastanglia.com", "DDD East Anglia"),
```
adrianbanks, line 30: Should this test be asserting the name and email address that the message got sent from?

DDDEastAnglia/Services/Messenger/Email/MailMessage.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            unchecked
+            {
+                int hashCode = (From != null ? From.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (To != null ? To.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (Subject != null ? Subject.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (Body != null ? Body.GetHashCode() : 0);
+                return hashCode;
+            }
+        }
```
adrianbanks: Bleugh - can we get rid of the equality operators from this class?

adrianbanks:
We seem to have a very inconsistent style of tests. Some tests are fluently named, others (the majority) are called *Tests, then there are a few *Facts. In addition, Some_Tests_Use_This_Format WhileOthers_UseThis.

DDDEastAnglia.Tests/Helpers/Email/SendGrid/HtmlRendererFacts.cs
alastairs: I can see where you're coming from, but I generally find that I don't actually care about the specifics of the SUT in a test, what I care about is understanding what is fulfilling that role in the test, the same way I care about the expected and actual results. Here, all I really care about is obtaining an instance of the SUT.

DDDEastAnglia.Tests/Services/Messengers/Email/EmailMessengerFacts.cs
alastairs, line 30: Assuming we care about that (which I think we do), then yes. Otherwise we can remove it from the equality calculation.

DDDEastAnglia/Services/Messenger/Email/MailMessage.cs
alastairs: No, see the commit history. In order for NSubstitute to be able to assert that a particular `MailMessage` was delivered, we need to be able to compare two instances of `MailMessage` for equality. It seems that NSubstitute doesn't support alternative methods of equality comparison, so we have to go with this.

It's not that nice, I agree, but I think we're stuck with it.

DDDEastAnglia.Tests/Helpers/Email/SendGrid/HtmlRendererFacts.cs
adrianbanks: I just find it _very_ distracting. I read it as "do something, but I don't know what, then use the thing it returned that I don't know what it is and call a method `Render` on it". All of the vagueness means I find it harder to reason about the test.

In some other tests below, you've named the variable something related to what it is and that is less distracting.

DDDEastAnglia.Tests/Services/Messengers/Email/EmailMessengerFacts.cs
adrianbanks, line 30: Feels like we should have a test for "all emails come from admin@dddeastanglie.com", which isn't this test.

DDDEastAnglia/Emails.resx
```diff
@@ -0,0 +1,144 @@
...
+        
+If you made this request, then please visit the address below.
+       
+[resetLink]
+
+This reset request will expire after 2 hours.
+                
+If you did not ask to change your password, then please ignore this email. Another user may have entered your username by mistake. No changes will be made to your account.
+
+You can contact us at organisers@dddeastanglia.com.</value>
```
adrianbanks, line 131: This bit of text was only in the text email, not the HTML email, as that is already present in the HTML template. We now get something that reads a bit odd.

DDDEastAnglia/Emails.resx
```diff
@@ -0,0 +1,144 @@
...
+    <comment>Sent when a password reset request is received for an account.</comment>
+  </data>
+  <data name="SessionSubmission" xml:space="preserve">
+    <value>Thanks for submitting a session for DDD East Anglia!
+        
+Here's a copy of your session abstract for your records:
+Title: [SessionTitle]
+[SessionAbstract]
+
+You can contact us at organisers@dddeastanglia.com.</value>
```
adrianbanks, line 141: Same here.

alastairs:
> We seem to have a very inconsistent style of tests. Some tests are fluently named, others (the majority) are called *Tests, then there are a few *Facts. In addition, Some_Tests_Use_This_Format WhileOthers_UseThis.

Yes. This is the case all over the code base. When introducing new tests I have tried to remain consistent  with what is around them. If you can point to specific new test fixtures/classes that aren't consistent with their surroundings then I'm happy to change those (or feel free to update them yourself :smile:)

adrianbanks:
Because we are using the nested test fixture pattern, R# is showing several inconclusive test runs due to there being fixtures with no tests in them - annoying.

DDDEastAnglia/Services/Messenger/Email/MailMessage.cs
adrianbanks: I've found a way to do this that contains the equality logic to the tests only. [Take a look](https://github.com/adrianbanks/dddeastanglia/commit/ea977dc76a48d77c472353502e21560accbecbc1). I think it is a bit nicer.

alastairs: Yeah, that works for me. I'll refactor to this.

adrianbanks: You could just pull in that change, as I did it in a branch off this pull request.

alastairs:
> Because we are using the nested test fixture pattern, R# is showing several inconclusive test runs due to there being fixtures with no tests in them - annoying.

I'm not seeing this, before or after the commit removing the nested test fixture. I've seen it previously with other solutions (e.g. at work), and found that restarting R#/VS can help.

DDDEastAnglia/Emails.resx
alastairs, line 131: Does this only apply to the "You can contact us..." line? Annoying we didn't spot this. There might be a way to work it into the refactored code.

adrianbanks, line 131: Yes, just that line (in the HTML email it is in the footer).

adrianbanks:
> I've seen it previously with other solutions (e.g. at work), and found that restarting R#/VS can help.

Just fired up VS again and it is still happening. There are about 5 test fixtures that show up as inconclusive.

alastairs:
I think the only outstanding comment is the one about the "contact us" footer. I'll have a think about how to incorporate that into the current design. Off the top of my head I came up with the following options:
1. Hack it into the `SendGridPostman` with string concatenation
2. Create another `IRenderer` for plain-text rendering. This doesn't feel quite right either - we're not rendering the plain-text mail, we just want to include some extra stuff in it. It's also a bit weird to pass two different implementations of an interface into a class

Taking option 1 might be the pragmatic approach, and I'd wrap the implementation in a private method so the responsibility can be moved elsewhere relatively easy in future. 

I'll sleep on it and see if anything better comes to mind tomorrow.

philpursglove:
Works on my machine™ (where 'Works' == 'actually it fails, but in the expected places')

DDDEastAnglia.Tests/Services/Messengers/Email/EmailMessengerFacts.cs
alastairs, line 30: It _is_ this test :smile: This is a collaboration test*, ensuring the `EmailMessenger` calls the `IPostman` correctly: is the correct method on `IPostman` called, and is it passed the correct arguments?

The `EmailMessenger` is responsible for preparing a `MailMessage` to be passed to its `IPostman` collaborator. There will only be this one implementation of `EmailMessenger`, even if we were to switch to something other than SendGrid as that responsibility is abstracted by the `IPostman`. The test asserts that `IPostman.Deliver()` is called with the expected `MailMessage` instance; that instance includes the `From` address details. Therefore, the test does correctly assert that all emails are sent from "DDD East Anglia <admin@dddeastanglia.com>".

\* See [Integration Tests are a Scam](https://vimeo.com/80533536) by J.B. Rainsberger, _well_ worth a watch.

philpursglove:
:shipit: This needs to be merged and in production in time for session submissions opening on Monday. Can we go with it as is and fix the duplication afterwards?

adrianbanks:
> This needs to be merged and in production in time for session submissions opening on Monday. Can we go with it as is and fix the duplication afterwards?

Yes. The problem with the duplicate link is (I think) down to trying to deal with HTML and text emails in the same manner. That can be addressed later though - I've added it as an issue (#335).

PR #337, opened 2015-05-31
Fix #333: Fix google logins

### LINKED ISSUE #333: Logging in with Google no longer works
When attempting to log in, the user gets redirected to this page:

https://support.google.com/accounts/answer/6206245?p=openid&rd=1

adrianbanks:
This should fix the Google logins after their upgrade to OAuth2.

I've got round the problem of upgrading existing pre-OAuth2 logins by simply removing them - they are easy enough to recreate.


adrianbanks:
Just tested this and it seems to work.

PR #349, opened 2017-05-04
Upgrade to Bootstrap 4

adsurg:


alastairs (commented):
Would be worth double-checking the Views and Partials to make sure there's nothing further to modify. Also, have you tested the Admin area with these changes in place?

adsurg:
Will do

adsurg:
@alastairs the Account pages appear to be missing from the live site, so I can't compare them to the code I'm looking at here.  The register page is showing a blue button which I assume is incorrect.  The forms are using Bootstrap default layouts and using a control-group class that's no longer there.  I wonder if I need a slightly better understanding of the structure of the site to finish this up, but in any case, it doesn't look like very much work to get this finished.

alastairs:
You might have more luck comparing against http://dddeastaging.azurewebsites.net/, although it looks like that's also in shutdown mode.

adrianbanks:
I've checked this in the main site and in the admin area. These are the issues I've found:

- The site no longer appear responsive. When shrunk down to phone size, the menu bar no longer turns into the hamburger menu and looks terrible when it wraps.
- The log in button has lost its background colour, but it still looks ok.
- The spacing around the form entry boxes has gone missing.
- The admin area is *completely* broken - it looks like the menu has lost its styling entirely and overlaps the content so that you can't interact with it. The main content of the pages look ok though.

Other than those issues, it seems to be close.

adrianbanks:
Still not working I'm afraid. The default view looks like:

image

After pressing the toggle button:

image

It always looks like this, even when there is no need to hide the menu.

adrianbanks:
I'm going to close this as we are changing the layout soon anyway.

PR #353, opened 2017-05-16
Remove an unused project

adrianbanks:
We don't use this anywhere.


philpursglove:
Fine with me, I don't think I even knew we **had** a SQL project, I thought all the database changes were in the Migrations project.

adrianbanks:
Yeah, it's not included in the solution. Adam got confused by it when he was trying to get the code to build.

PR #364, opened 2018-06-14
Added notice about not being able to cover speaker expenses

philpursglove:


philpursglove:
@adrianbanks Can you see why the CI build failed?

adrianbanks:
I've tweaked this to an info alert instead of a well and merge it manually.

adrianbanks:
I'm not sure the integration is working properly. If it continues to not work, it is probably easier to just disable it for now.

PR #365, opened 2018-06-24
Commented out Google login as it's not currently working

philpursglove:
Intended as a short-term thing while we diagnose/fix the actual google login.

@adrianbanks I copied what you did for the Twitter login!


adrianbanks:
I can't reproduce this. It works for me and another account I tried, so not sure what the issue is.

I'm tempted to leave it there and see if there are any more complaints.

adrianbanks:
Superseded by #379.

PR #366, opened 2018-07-12
Show the number of distinct speakers in the summary at the top of the page

philpursglove:
Wanted to be able to see in one hit how many distinct speakers have submitted so we have a slightly better idea of when we have enough


adrianbanks:
Looks good. That had been niggling me as well. 👍

adrianbanks:
I think the AppVeyor build is not working properly.

PR #369, opened 2018-11-20
Add security headers

philpursglove:
Looking at our [SecurityHeaders report](https://securityheaders.com/?q=dddeastanglia.com&hide=on&followRedirects=on) we're currently getting an F - this PR is to add a number of the headers we're currently missing and also remove the ASP.NET version headers. Once this is merged I'll add issues for Content Security Policy/Feature Policy/Referer Policy.


adrianbanks:
Seems like a useful change. 👍

PR #377, opened 2019-05-24
Update carousel images, and start on the environmental policy

dcook-net:
Initial stab at the environmental page. Needs a little work in places. Feedback welcome.
Content not viewable to users yet.

I've updated the the images on the carousel to fresher and move diverse.
Logo on carousel is still the old green one, so that will still need changing


DDDEastAnglia/Controllers/HomeController.cs
```diff
@@ -56,6 +56,11 @@ public ActionResult CodeOfConduct()
         {
             return View();
         }
+        
+//        public ActionResult Environmental()
```
CraftyFella, line 60: This correct?

DDDEastAnglia/Controllers/NavigationBarController.cs
```diff
@@ -54,6 +54,7 @@ public ActionResult RenderMenu()
                     CreateLink("Accommodation", "Home", "Accommodation", conference.CanRegister),
                     CreateLink("Sponsors", "Home", "Sponsors"),
                     CreateLink("Team", "Home", "Team"),
+//                    CreateLink("Environmental Policy", "Home", "Environmental"),
```
CraftyFella, line 57: And this?

DDDEastAnglia/Controllers/HomeController.cs
dcook-net: yeah - still working on it and stupidly started the images on the same fork.
Images are higher priority for now. So keeping the Env stuff dark for now.

Little bit lazy - should have split it all up in to sep change sets!

DDDEastAnglia/Controllers/NavigationBarController.cs
dcook-net: as above

DDDEastAnglia/Controllers/HomeController.cs
CraftyFella: understood.. just checking it wasn't a mistake

PR #378, opened 2019-06-02
Add anonymous session submissions and voting

adrianbanks:
This adds the ability to make the whole session submission and voting process anonymous. When it is turned on, speaker details will not be shown alongside session details.

I've tested it in both session submission and voting stages and can't find any pages that still show the speaker details when the setting is turned on. They do still show the details when the setting is turned off.

### DIFF
--- a/DDDEastAnglia/Controllers/SessionController.cs
+++ b/DDDEastAnglia/Controllers/SessionController.cs
@@ -40,8 +40,7 @@ namespace DDDEastAnglia.Controllers
             var sessions = sessionRepository.GetAllSessions();
 
             var allSessions = new List<SessionDisplayModel>();
-
-            bool showSpeaker = conference.CanShowSpeakers();
+            var showSpeaker = conference.CanShowSpeakers();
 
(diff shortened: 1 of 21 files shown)


adrianbanks:
There are currently some failing unit tests on this branch.

DDDEastAnglia/Controllers/SessionController.cs
```diff
@@ -172,7 +178,7 @@ public ActionResult Delete(string userName, int id = 0)
             }
 
             var userProfile = userProfileRepository.GetUserProfileByUserName(session.SpeakerUserName);
-            var displayModel = CreateDisplayModel(session, userProfile);
+            var displayModel = CreateDisplayModel(session, userProfile, true);
```
dcook-net: hard coded true for showSpeaker here, rather than a toggle.
Assume this is intended?

adrianbanks: 👍 Good spot.

PR #380, opened 2019-06-06
Add speaker resources and updated content for 'submissions open'

dcook-net:


adrianbanks:
I've made some changes to this by moving the guidance to the top of the page where the session is submitted instead of the sessions list as I think it makes more sense to see it there.

adrianbanks:
I've merged this manually.

PR #381, opened 2019-06-13
Branding play

adsurg:
I've made some branding updates.  There's bound to be more to do, but this should be a help.


CraftyFella (approved):
TBH.. i say merge it to staging and lets have a looksee

adrianbanks:
I was going to have a quick test of it later on. I'll put any issues I find on here.

adrianbanks:
Generally looks good. There are a few things I found when testing it:

- The notification banner is still green:



- There is a white banner at the top of the page that shouldn't be there:



- The menu bar needs some kind of background colour to stop it washing out in the page, especially when on a page that doesn't select an item:



- The sidebar is too wide and the sponsor logos wrap (they're not quite the same height so it looks odd):



- The titles in the sidebar don't contrast enough with the background:



- The font for the normal text in the sidebar is a bit too small (e.g. the dates) - it might be a bit too small on the whole site. Might just be on a high DPI screen that this appears.

- There is an odd indentation on the dates in the sidebar:



- The active colour of the menu links is still green:



- There's an odd grey bar covering the username when the menu is dropped down:



- The log off link is not aligned properly:



- There's a shadow on the text in the alerts, making the text look blurry:



- The other favicons (`Images/logo-icon-*.png`) need updating (these are for bookmarks on mobile devices):

adrianbanks:
I've merged this manually.

PR #382, opened 2019-06-18
The site should be more progressive.

adsurg:


adrianbanks:
Good work on hiding the conference name when the screen is small.

The sidebar seems to get very narrow when the screen is small:

image

It also is jumps about in size and position on the other pages - there doesn't seem to be any consistency as to why.

The sponsor logos have also lost their backgrounds and borders:

image

The width where the menu appears instead of individual links probably needs tweaking as well:

image

adrianbanks:
Merged this manually.

PR #383, opened 2019-06-20
Add session duration

adsurg:


adrianbanks:
Looks like you've forgot to check in the migration file.

adrianbanks:
I've merged this manually after rebasing it.

PR #385, opened 2019-07-26
Add duration icons to voting leaderboard

philpursglove:
Thought it would be useful to have the duration icons displayed on the leaderboard. Bonus points if anyone can tell me why those old commits are showing in this PR...


adrianbanks:
Did you open the PR before rebasing it? If so, that's why the commits are showing.

*EDIT* It looks like you've merged staging into a different branch that you've based this off. Not sure why the commits are showing but not the contents.

adrianbanks:
Merged this manually to avoid the unwanted commits.

